PR #5, opened 2018-09-03
Secure UI rules

tembleking:
Added support for the following Secure UI rules via Terraform rules and some examples:

- Processes
- Containers
- Network
- Filesystem
- Syscalls


nestorsalceda:
Looks good to me!

Thanks!

PR #6, opened 2018-10-01
Add Go 1.11 Modules support

tembleking:
Go 1.11 added a preliminary [support for modules](https://github.com/golang/go/wiki/Modules) and although they are experimental right now, [will continue to work with 1.12 and later](https://golang.org/doc/go1.11#modules).

They improve reliable and reproducible builds and replace tools like [dep](https://github.com/golang/dep) and [vgo](https://github.com/golang/vgo).

This PR just adds the go.mod and go.sum files to support working outside the GOPATH, help to communicate to the ecosystem that we are opting in to modules, and declare the identification of our code.


nestorsalceda:
This looks awesome. When golang 1.11 enters in Debian I will make the switch officially :D

Thanks you very much!

PR #12, opened 2020-03-02
Add sysdig_user resource

rung:
Signed-off-by: Hiroki Suezawa <suezawa@gmail.com>

### What
- Add sysdig_user resource (Sorry for a big PR)

### Why
- This PR is related to #11 
  - I would like to add `sysdig_secure_team` too, but this PR has `user` only.

### Special notes for your reviewer:
- I used `terraform-test@sysdig.com` for testing for now. but Does Sysdig have e-mail address for testing?

PR #14, opened 2020-03-06
Modify test for sysdig_user

rung:
### What
- change `resource_sysdig_user_test.go`
  - Doesn't change email address in each step. 

### Why
- test on #12 was failed.
  - (I thought it passed once, but it failed today. sorry.)


tembleking:
I've tried it and it's working, merging. Thanks for your work @rung  :slightly_smiling_face:

PR #31, opened 2020-05-29
add support for teams in Monitor

dprosper:
Here are the updates I feel I need for supporting teams in Monitor, unfortunately, I have no way of testing this with Secure to confirm those changes also work with Secure.  Can you please review, let me know if you have any question and consider implementing these after some testing?

### DIFF
--- a/sysdig/common/client.go
+++ b/sysdig/common/client.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"strings"
 	"crypto/tls"
 	"io"
 	"log"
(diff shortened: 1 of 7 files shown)


sysdig/common/client.go
```diff
@@ -36,10 +42,28 @@ type sysdigCommonClient struct {
 }
 
 func (client *sysdigCommonClient) doSysdigCommonRequest(method string, url string, payload io.Reader) (*http.Response, error) {
-	request, _ := http.NewRequest(method, url, payload)
-	request.Header.Set("Authorization", "Bearer "+client.SysdigAPIToken)
+
+	var IBMInstanceID string
```
nestorsalceda, line 46: I would prefer to not have IBM platform specific code here.

Is a pretty good concern and I have to ask about if we need to support IBM platform in addition to vanilla Sysdig platform. I'm going to raise this concern internally.

On any case, we need to think how to keep this open to allow others to implement these kinds of personalization.

sysdig/common/client.go
```diff
@@ -13,6 +14,11 @@ type SysdigCommonClient interface {
 	GetUserById(int) (User, error)
 	DeleteUser(int) error
 	UpdateUser(User) (User, error)
+
+	CreateTeam(Team) (Team, error)
```
nestorsalceda, line 18: Teams in monitor are not the same teams that ones in secure so that this should not be in Common.

@tembleking implemented teams support for monitor in: https://github.com/draios/terraform-provider-sysdig/commit/1de6ebf2df9b68bf34b745dcf7db7b814ef3925a

PR #39, opened 2020-08-07
feat: User password

tembleking:
This PR adds the password field to the user resource so the user does not get an invitation e-mail to create the account. The defined password is the password at the moment of the creation and will not be validated again if the user changes it.


tembleking:
This PR is a draft and is prone to be closed without merging it, since the API ignores the `password` field in the requests.
Validating this behavior with the backend team.

PR #47, opened 2020-09-29
refactor: Move creation logics from Dashboard v3 to the client

tembleking:
This PR pushes the creation logic of Dashboards, Panels and Queries to the client


nestorsalceda (approved):
LGTM, thank you!

PR #79, opened 2021-01-13
feat: Vulnerability exception resources

tembleking:
This PR adds the `sysdig_secure_vulnerability_exception` and `sysdig_secure_vulnerability_exception_list` resources, to manage the Vulnerability exceptions as code.

### DIFF
--- a/sysdig/internal/client/secure/client.go
+++ b/sysdig/internal/client/secure/client.go
@@ -48,6 +48,10 @@ type SysdigSecureClient interface {
 	DeleteVulnerabilityExceptionList(context.Context, string) error
 	UpdateVulnerabilityExceptionList(context.Context, *VulnerabilityExceptionList) (*VulnerabilityExceptionList, error)
 
+	CreateVulnerabilityException(context.Context, string, *VulnerabilityException) (*VulnerabilityException, error)
+	GetVulnerabilityExceptionByID(context.Context, string, string) (*VulnerabilityException, error)
+	DeleteVulnerabilityException(context.Context, string, string) error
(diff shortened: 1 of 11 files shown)


JordanSussman:
Thanks @tembleking! 

Any idea if the API team eventually has plans to add some endpoint to manage a list of exceptions with a single API call instead of one per cve? I have some worries that managing the 3,000 exceptions or so I have will cause lengthy plan/apply times and/or rate limits.

tembleking:
Thank you for the feedback @JordanSussman :+1:  users always provide the most valuable information for the project :slightly_smiling_face: 

Regarding the API calls, the method that retrieves the info from the Vulnerability Exception List contains the full list of exceptions, so if in the future the Read method needs to be optimized, we can use that. I don't think the API rate limits will be something to worry about anymore since the retry strategy was implemented (#65), it will try again after some time if the API complains. Also, Terraform allows to throttle the number of API calls performed in parallel with [`--parallelism`](https://www.terraform.io/docs/commands/apply.html#parallelism-n) which by default is 10.
If there is a lot of infra to manage via Terraform, yes, this will cause a lengthy plan. Are you currently hitting limits? If so, please let me know so we can plan how to solve it.

JordanSussman:
> If there is a lot of infra to manage via Terraform, yes, this will cause a lengthy plan. Are you currently hitting limits? If so, please let me know so we can plan how to solve it.

Our immediate plan is to separate the plan/apply to only target the file(s) changed within GitHub to reduce the overall timeline. I'll let you know if we run into issues with limits.

Thanks again!

PR #82, opened 2021-01-28
Let's assume there's no whitespace at all

blacksd:
Since the API allow storing the field with or without whitespace (or in a fancy combination), let's just assume there's no whitespace at all and act on that.
Or, we could use regexp and have a more robust solution.


blacksd:
Closing this as superseded by #83

PR #116, opened 2021-08-18
Properly unmarshal trustedIdentity response

nkraemer-sysdig:
The trustedIdentity datasource was not properly unmarshalling the string response from the backend, resulting in extra quotations that broke the consumer. 

Add a fix to properly unmarshal this response.


wideawakening (approved):
tested provider locally with aws module and provisioning went without errors this time :+1:

nestorsalceda (approved):

PR #120, opened 2021-09-01
Add a data source to wrap Fargate tasks in the Sysdig workload agent

gnosek:
This is heavily based on https://github.com/falcosecurity/kilt/tree/master/runtimes/terraform but comes with a built in definition for the Sysdig Fargate workload agent


sysdig/data_source_sysdig_fargate_workload_agent.go
```diff
@@ -0,0 +1,208 @@
...
+
+const agentinoKiltDefinition = `build {
+    entry_point: ["/opt/draios/bin/instrument"]
+    command: ${?original.entry_point} ${?original.command}
+    environment_variables: {
+        "SYSDIG_ORCHESTRATOR": ${config.orchestrator_host}
+        "SYSDIG_ORCHESTRATOR_PORT": ${config.orchestrator_port}
+        "SYSDIG_COLLECTOR": ${config.collector_host}
+        "SYSDIG_COLLECTOR_PORT": ${config.collector_port}
+        "SYSDIG_ACCESS_KEY": ${config.sysdig_access_key}
```
balinaveen, line 22: Eric Lugo and I briefly looked at this. Does it make sense to move these env variables to be part of the top level provider config where we currently store the api tokens for monitor and secure?

gnosek: No idea. Right now if you want to provision multiple agentinos with different config, you get that for free (as you have to put these params in the data source). If we move these to the provider config, we'll have to set up multiple providers if we want different agentino configs. I'm not sure if terraform modules can easily create new provider instances (they probably can but there was something not 100% obvious about it)

go.mod
```diff
@@ -7,7 +7,9 @@ require (
 	cloud.google.com/go/storage v1.13.0 // indirect
 	github.com/agext/levenshtein v1.2.3 // indirect
 	github.com/apparentlymart/go-cidr v1.1.0 // indirect
-	github.com/aws/aws-sdk-go v1.37.10 // indirect
+	github.com/aws/aws-sdk-go v1.37.10
```
admiral0, line 10: is there a reason why we aren't using sdk v2 for aws?

sysdig/data_source_sysdig_fargate_workload_agent.go
```diff
@@ -0,0 +1,208 @@
+package sysdig
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"errors"
+	"fmt"
```
admiral0, line 8: leave an empty line after the stdlib deps

sysdig/data_source_sysdig_fargate_workload_agent.go
admiral0: these settings are per VPC we want to monitor, so they need to be more specific

sysdig/data_source_sysdig_fargate_workload_agent.go
```diff
@@ -0,0 +1,208 @@
...
+			"workload_agent_image": {
+				Type: schema.TypeString,
+				Description: "the Sysdig workload agent image",
+				Required: true,
+			},
+			"image_auth_secret": {
+				Type: schema.TypeString,
+				Description: "registry authentication secret",
+				Optional: true,
+			},
```
admiral0, line 59: this will be unused as our images are currently downloadable without auth

sysdig/provider.go
```diff
@@ -93,6 +93,8 @@ func Provider() *schema.Provider {
 			"sysdig_secure_notification_channel":   dataSourceSysdigSecureNotificationChannel(),
 			"sysdig_current_user":                  dataSourceSysdigCurrentUser(),
 			"sysdig_user":                          dataSourceSysdigUser(),
+
+			"sysdig_fargate_workload_agent": dataSourceSysdigFargateWorkloadAgent(),
```
admiral0, line 97: I would name it `sysdig_instrumented_task_definition`

admiral0 (approved):
some minor issues

sysdig/data_source_sysdig_fargate_workload_agent.go
0snug0: What about the Access Collector and Access Key? Those will not change

go.mod
gnosek: No idea, it was already there apparently 🤷‍♂️

sysdig/data_source_sysdig_fargate_workload_agent.go
gnosek: I expect this is for people mirroring the image to their own private repo that might need authentication?

sysdig/data_source_sysdig_fargate_workload_agent.go
gnosek: I'm not convinced splitting the settings across the provider/data source will be more intuitive but if you feel strongly about it, just tell me the list of settings you want moved to the provider

PR #123, opened 2021-09-07
Allow type to be specified for alert resources

jon-ruckwood:
Default to `MANUAL` when not set explicitly.

Type `PROMETHEUS` can be used to specify an alert with a PromQL expression.

Issue #108, opened 2021-07-14
Support for PromQL Alerts


stuwatts:
We're starting to make more use of PromQL based alerts.  I appreciate they're in Beta at the moment, but it would be great to get support for these in the TF provider.  We can't adopt the provider until PromQL alerts are supported.

tembleking:
Thanks @stuwatts for the feedback, let me talk with the team and prioritize this Feature Request :slightly_smiling_face:

allen-servedio:
Hi @tembleking - we have also hit a point where we could make use of this feature. Has your team decided if they will be taking this up and - if so - any rough idea on when? Thanks either way!

jon-ruckwood:
Possibly related to https://github.com/sysdiglabs/terraform-provider-sysdig/pull/123

DenysChekirda:
Fix proposed by @jon-ruckwood is working in case of this issue. I've been able to create a prometheus alert.

tembleking:
This can be closed now, since it has been implemented by #129

PR #129, opened 2021-09-30
feat: Support for PromQL Alerts (#108)

dbonf:
This adds support to PromQL alerts using the new resource: `sysdig_monitor_alert_promql`.

The same could be achieved in a more succinct way defining a new field in the resource `sysdig_monitor_alert_metric` as described by the good PR https://github.com/sysdiglabs/terraform-provider-sysdig/pull/123, reported here is an alternative way: my understanding is that a new resource could provide a more coherent interface from a user perspective.


tembleking:
I agree that creating a new resource would provide a better interface to final users and I also think they would allow both resources to grow at different rates, or implement different features, as the product evolves.

Thank you so much for the contribution @dbonf :100: LGTM

PR #133, opened 2021-10-11
Parse Azure trusted identity

aq17:
Split trusted identity into tenantID/ clientID pair, needed to authenticate the service principal running cloud benchmarks


wideawakening:
merged in #134

PR #140, opened 2021-11-02
Refactor Azure Service Principal Attributes

aq17:
What was previously referred to as the client ID is actually the service principal ID; client ID should represent the ID of the application in which the service principal lives, and is not needed by the TF provider.


sysdig/data_source_sysdig_secure_trusted_cloud_identity.go
```diff
@@ -43,7 +43,7 @@ func dataSourceSysdigSecureTrustedCloudIdentity() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"azure_client_id": {
+			"azure_service_principal_id": {
```
nkraemer-sysdig, line 46: This will need a corresponding update in the azure module

nkraemer-sysdig (approved):

PR #148, opened 2022-01-07
Update to compliance v2 task

aq17:
Change terraform deploys to create compliance v2 task instead of benchmarks v2 task

https://sysdig.atlassian.net/browse/SSPROD-10400


nkraemer-sysdig (commented):
Doesn't the model need to change as well? The request/response bodies are different between the two endpoints aren't they?

aq17:
> Doesn't the model need to change as well? The request/response bodies are different between the two endpoints aren't they?

Don't think so, looks like the new endpoint did a good job of adhering to the old schema for benchmarks (going based off the docs)

nkraemer-sysdig (approved):

aq17:
@nestorsalceda @tembleking @wideawakening would any of you be familiar with the failing acceptance tests?

nestorsalceda:
I'm checking with @tembleking and we think it might be a dependencies problem.

We are going to update and rerun this check

Thank you folks!

nestorsalceda:
It worked, please merge if you need to :)

PR #149, opened 2022-01-11
Respect proxy env vars in common HTTP client

chrisob:
Currently, API calls for "Sysdig Platform" resources/data sources from the "common" client don't respect proxy env vars (HTTPS_PROXY, NO_PROXY, etc.). When using the provider behind a proxy, this results in what seems like a provider hang.

While the monitor & secure clients do work behind a proxy & respect proxy env vars, any management of platform resources (ie. `resource "sysdig_user"`, `data "sysdig_current_user"`) fails behind a proxy.

This PR configures the common client to respect proxy settings like the [monitor](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/master/sysdig/internal/client/monitor/client.go#L50) & [secure](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/master/sysdig/internal/client/secure/client.go#L79) clients already do.

### DIFF
--- a/sysdig/internal/client/common/client.go
+++ b/sysdig/internal/client/common/client.go
@@ -31,6 +31,7 @@ func NewSysdigCommonClient(sysdigAPIToken string, url string, insecure bool) Sys
 	client.HTTPClient = &http.Client{
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure},
+			Proxy:           http.ProxyFromEnvironment,
 		},
 	}
 
--- a/sysdig/internal/client/common/client.go
+++ b/sysdig/internal/client/common/client.go
@@ -28,12 +28,12 @@ func WithExtraHeaders(client SysdigCommonClient, extraHeaders map[string]string)
 
 func NewSysdigCommonClient(sysdigAPIToken string, url string, insecure bool) SysdigCommonClient {
 	client := retryablehttp.NewClient()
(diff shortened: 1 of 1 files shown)


chrisob:
:wave:  @nkraemer-sysdig @tembleking 

Our BigCorp just integrated Sysdig, and we'd love to use the TF provider to manage our config. It'd be great to see this fixed & released soon so we can manage our users via code :grin:

airadier:
I would recommend the approach of cloning the Default Transport instead:

```
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.TLSClientConfig = TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure}
```

this will keep the sane default configs from the http.DefaultTransport, including honoring the proxy settings.

chrisob:
@airadier Thanks for the suggestion :+1: implemented & rebased on the latest master

nkraemer-sysdig (approved):
LGTM

chrisob:
Thanks for the approval, but it looks like some CI workflows still need to be manually approved before running:

image

nkraemer-sysdig:
Hey @chrisob, thanks for the PR! I ran the workflow, but since your fork does not have access to the secrets in the sysdiglabs repo, the tests won't be able to run. I have cloned this PR in the upstream repo in https://github.com/sysdiglabs/terraform-provider-sysdig/pull/151

tembleking:
Thanks for the PR @chrisob!

PR #152, opened 2022-02-04
Update cloud account URL to upsert

aq17:
URL needs to include `upsert` query param to upsert a cloud account if it already exists, avoiding 409 conflict


wideawakening:
~long~ short live 409!

aq17:
@wideawakening , can we actually confirm first if the management cloud account is already created in whatever instance the test suit runs on? It may be possible that the management cloud account was previously created and never destroyed, which would explain the conflict. Also we may want to confirm that nothing in the test suite depends on the management cloud account always existing, though that seems unlikely
cc @nkraemer-sysdig

PR #157, opened 2022-02-15
feat: add datasource secure connection

wideawakening:
in order terraform modules to be able to infer the secure endpoint/apiKey from the provider.
this will deprecate current module input vars

### DIFF
--- a/sysdig/data_source_sysdig_secure_current_connection.go
+++ b/sysdig/data_source_sysdig_secure_current_connection.go
@@ -15,12 +15,12 @@ func dataSourceSysdigSecureConnection() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"secure_url": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Compute:     true,
 				Description: "Sysdig Secure URL basepath to where backend requests will be sent",
 			},
 			"secure_api_token": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Compute:     true,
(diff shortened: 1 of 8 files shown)


sysdig/sysdig_clients.go
```diff
@@ -12,6 +12,9 @@ import (
 )
 
 type SysdigClients interface {
+	GetSecureEndpoint() (string, error)
```
wideawakening, line 15: dunno if this is the best place from where to fetch the two attributes.
accept any other suggestion

sysdig/sysdig_clients.go
```diff
@@ -12,6 +12,9 @@ import (
...
+	GetSecureApiToken() (string, error)
```
wideawakening, line 16: dunno if this is the best place from where to fetch the two attributes.
accept any other suggestion

sysdig/data_source_sysdig_secure_current_connection.go
```diff
@@ -0,0 +1,55 @@
...
+		Schema: map[string]*schema.Schema{
+			"secure_url": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Sysdig Secure URL basepath to where backend requests will be sent",
+			},
+			"secure_api_token": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Sensitive:   true,
```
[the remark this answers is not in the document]
wideawakening: reverted to `Compute` as it makes more sense

PR #168, opened 2022-04-15
Support auto creation of missing benchmark resources (SSPROD-12557)

aq17:
Other TF providers create a resource if it is missing (i.e. was manually deleted) instead of throwing a 404

Jira ticket: https://sysdig.atlassian.net/browse/SSPROD-12557

### DIFF
--- a/sysdig/resource_sysdig_secure_benchmark_task.go
+++ b/sysdig/resource_sysdig_secure_benchmark_task.go
@@ -3,6 +3,7 @@ package sysdig
 import (
 	"context"
 	"strconv"
+	"strings"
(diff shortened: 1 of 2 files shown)


nkraemer-sysdig (approved):

aq17:
@wideawakening could you please take a look at the failing test 😄

wideawakening:
failing tests have nothing to do with your changes :) 

- looks like something was changed on back-end error payload. team has been giving it a look at #170 
- we updated monitor api token too, as it was deprecated
- still with team but still, we're dealing with `429` throttling errors.

PR #175, opened 2022-06-02
ECS json bug fix that corrects casing issues for entrypoint, image an…

yashgiri:
…d environment values


tembleking:
Redirecting review to @gnosek because he implemented the fargate data source, and I don't have context on this

sysdig/template.go
```diff
@@ -0,0 +1,97 @@
+package sysdig
```
achandras, line 1: this is a pretty generic filename, could it be called something more descriptive like `cfn_preprocess_template.go`?

a comment block at the top of this file explaining the high level purpose of these functions would be useful as well

sysdig/testfiles/ECSinput.json
```diff
@@ -0,0 +1,30 @@
+[
```
achandras, line 1: This is great! I think a few more test files would be useful but that can be added in the future, since this looks like it's exercising the paths we're worried about

sysdig/cfn_preprocess_template.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// This code block is used when parsing ECS JSON format
+	for _, resource := range template.S("Resources").ChildrenMap() {
+		for _, container := range resource.S("Properties", "ContainerDefinitions").Children() {
+			if container.Exists("image") {
+				passthrough, _ := GetValueFromTemplate(container.S("image"))
+				_, err = container.Set(passthrough, "Image")
+				if err != nil {
+					return nil, fmt.Errorf("Could not append entrypoint values to EntryPoint: %v", err)
```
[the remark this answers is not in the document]
yashgiri: You are correct, I will fix this.

achandras (approved):

PR #177, opened 2022-07-08
fix: json casing

achandras:
https://sysdig.atlassian.net/browse/SMAGENT-3875

If there are keys that we know to be used by Kilt, those will be converted to Cloudformation JSON syntax, which essentially just means capitalized keys names, before being passed to the patcher.

Also adds some more test coverage specifically around the conversion.


marojor (approved):
LGTM

PR #173, opened 2022-05-18
feat: add alert group name

arturodilecce:
Adds the possibility to specify a `group_name` attribute on all Monitor Alert resources. This helps in the UI to group together similar alerts.

PR #181, opened 2022-07-30
feat: add alert group name v2

hrabalvojta:
Add new field to monitor, to help organize alerts in sysdig UI. Instead of having them all in default.


hrabalvojta:
Previous version #173 did not work. And contained blocking elements.

arturodilecce (approved):

arturodilecce (commented):
I've fixed https://github.com/sysdiglabs/terraform-provider-sysdig/pull/173 and also tested it. Seems to be working.

PR #186, opened 2022-08-08
fix: add the parameter sysdig_logging to the fargate_workload_agent

francesco-racciatti:
This PR adds the parameter `sysdig_logging` to the `fargate_workload_agent` so that users can set the instrumentation logging level at deployment time.

### DIFF
--- a/.github/workflows/ci-pull-request.yml
+++ b/.github/workflows/ci-pull-request.yml
@@ -75,6 +75,7 @@ jobs:
       - name: Get dependencies
         run: |
           go get -u -v github.com/onsi/ginkgo/ginkgo
+          go mod tidy
 
       - name: Test
         run: make test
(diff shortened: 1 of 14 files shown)


francesco-racciatti:
@achandras @yashgiri 
Hey folks, 
https://sysdig.atlassian.net/browse/SSPROD-14387 prevents custom rules from having the same name as existing default rules. This is making the automated test suite scream.

In 2735cb6a4456750ba14500ec8e75dcb6777dbbfa I disabled the test that performs the banned action (`custom rules name = default rule name`).

But I’m not familiar with this test codebase, so I don’t know if this is the right way to cope with that.
What do you think about it?

achandras:
I am not sure either, @tembleking do you have any thoughts on this?

achandras (approved):

PR #188, opened 2022-08-16
feat: configure Fargate instrumentation log group

achandras:
Adds a new optional parameter block `log_configuration` which can be used to configure the `LogConfiguration` block on the sysdig instrumentation container, using the `awslogs` driver.


francesco-racciatti (approved):
LGTM

yashgiri (approved):

PR #189, opened 2022-08-23
ci: Update setup-go action

tembleking:


wideawakening (approved):
thanks for the maintenance 👍🏽

PR #198, opened 2022-09-09
doc: remove documentation about direct mode

marojor:
There's no benefit in running direct mode, so remove the documentation for it. The right way of connecting to the backend is via the orchestrator agent.


balinaveen (approved):
LGTM

PR #205, opened 2022-10-31
fix: add name as required param of capture action

rbaderts:
Please review.   I figured out that a previous test run had leaked the test sysdig user resource causing subsequent acc test runs to fail.... I  will submit a PR separately which adds a Sweeper to resource_sysdig_user_test to account for this happening, but I imagine it could be a general problem and that all the test cases need this...


sysdig/resource_sysdig_secure_policy.go
```diff
@@ -107,6 +107,10 @@ func resourceSysdigSecurePolicy() *schema.Resource {
 										Required:         true,
 										ValidateDiagFunc: validateDiagFunc(validation.IntAtLeast(0)),
 									},
+									"name": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
```
tembleking, line 113: By marking the new field as required, this is a breaking change for the users. They may not be expecting the need to add this kind of field. Either we make it optional, and provide a default value, or we release a new breaking Semver.

wideawakening:
please document new attributes on documentation, as it's not generated automagically 🙏🏽 
`/r/website/docs/r/secure_policy.md`

sysdig/resource_sysdig_secure_policy.go
```diff
@@ -107,6 +107,11 @@ func resourceSysdigSecurePolicy() *schema.Resource {
 										Required:         true,
 										ValidateDiagFunc: validateDiagFunc(validation.IntAtLeast(0)),
 									},
+									"name": {
+										Type:     schema.TypeString,
+										Optional: true,
+										Computed: true,
```
tembleking, line 113: AFAIK the value can be either Optional or Computed, but not both. If the user is able to specify it, it has to be Optional, if the value is filled by the provider itself (as an output parameter), then it needs to be Computed.

sysdig/resource_sysdig_secure_policy.go
```diff
@@ -225,12 +231,23 @@ func addActionsToPolicy(d *schema.ResourceData, policy *secure.Policy) {
...
 		afterEventNs := d.Get("actions.0.capture.0.seconds_after_event").(int) * 1000000000
 		beforeEventNs := d.Get("actions.0.capture.0.seconds_before_event").(int) * 1000000000
-		policy.Actions = append(policy.Actions, secure.Action{
-			Type:                 "POLICY_ACTION_CAPTURE",
-			IsLimitedToContainer: false,
-			AfterEventNs:         afterEventNs,
-			BeforeEventNs:        beforeEventNs,
-		})
+		name := d.Get("actions.0.capture.0.name").(string)
+		if len(name) <= 0 {
```
tembleking, line 235: Let's use `if name == "" {` instead to check if the value is empty.

sysdig/resource_sysdig_secure_policy.go
```diff
@@ -225,12 +231,23 @@ func addActionsToPolicy(d *schema.ResourceData, policy *secure.Policy) {
...
+			})
+		} else {
+			policy.Actions = append(policy.Actions, secure.Action{
+				Type:                 "POLICY_ACTION_CAPTURE",
+				IsLimitedToContainer: false,
+				AfterEventNs:         afterEventNs,
+				BeforeEventNs:        beforeEventNs,
+				Name:                 name,
+			})
+		}
```
tembleking, line 250: Both the if and the else are similar in case the Name is `""`, you can get rid of the if/else and just use the append.
```suggestion
		policy.Actions = append(policy.Actions, secure.Action{
			Type:                 "POLICY_ACTION_CAPTURE",
			IsLimitedToContainer: false,
			AfterEventNs:         afterEventNs,
			BeforeEventNs:        beforeEventNs,
			Name:                 name,
		})
```

tembleking (changes requested):

sysdig/resource_sysdig_secure_policy.go
rbaderts: Apparently this is not the case.   When both a property is declared Optional and Computed, the value is only computed when not provided.

https://github.com/hashicorp/terraform/issues/21278

sysdig/resource_sysdig_secure_policy.go
rbaderts: just committed your proposed change which addressed this and the previous comment

ivanlysiuk-sysdig:
@tembleking ,
are there remaining blockers on this PR to get it merged?

sysdig/resource_sysdig_secure_policy.go
tembleking: Didn't know that, thanks for pointing out

tembleking (approved):
LGTM, thanks!

hayk99 (approved):
🆗

miketnt (approved):

PR #206, opened 2022-11-01
added sweepers to cleanup leaked test user resource

rbaderts:
For whatever reason the testing user was leaked by previous acc test runs.....   Once in this state subsequent ACC tests will fail....   Sweepers are a terraform SDK concept for cleaning this stuff up....   Not sure if the "checkDestroy" facility might be a better solution...

Also includes changes to github lint workflow to get around a deprecation warning.   I'm not sure what triggered the deprecation warning, but upgrading golangci/golangci-lint-action to v3 seems to have fixed it...


sysdig/data_source_sysdig_user_test.go
```diff
@@ -36,13 +36,13 @@ func init() {
 			ctx := context.Background()
 			user, err := commonClient.GetUserByEmail(ctx, "terraform-test+user@sysdig.com")
 
-			if err != nil {
-				return err
+			if user == nil || err != nil {
+				return nil
 			}
```
tembleking, line 41: In case of error we cannot return a nil, if there's an error we need to inform the user of the provider that something wrong happened.

tembleking: I am seeing this same pattern everywhere, we need to return the error to the user of the provider, otherwise, they wont have feedback and the infrastructure will remain in an inconsistent state, because terraform will store this state but the infra is actually not modified accordingly.

tembleking (changes requested):

sysdig/data_source_sysdig_user_test.go
rbaderts: Ya I agree, I will need to change the function to not return an error when a user with that email is not found, because the user may or may not exist when this sweep() function is invoked, and it's not an error if it doesn't...

rbaderts: fixed.   Also fixed getUser to pass in a query param rather than getting them all from the server and then filtering....

wideawakening:
thx Rick :) 
reopening to review if any of these changes make sense (some i do think they're worthy keeping)

Issue #223, opened 2022-12-12
Provider v0.5.43 crashes after appending exception to managed rule


omaen:
Hi,

I tried the new version that would allow the creation of custom exceptions to existing managed rules in Sysdig Secure. The creation of these rules were successful, but the next time I tried to do ta `terraform plan` while working on a standalone custom rule, the provider crashes with this exception message:
```
The plugin encountered an error, and failed to respond to the
plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more
details.

Stack trace from the terraform-provider-sysdig_v0.5.43 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc743ed]

goroutine 42 [running]:
github.com/draios/terraform-provider-sysdig/sysdig.resourceSysdigRuleFalcoRead({0x10a7328, 0xc0006e14a0}, 0xc0006e8900, {0xd76da0?, 0xc000598690})
	github.com/draios/terraform-provider-sysdig/sysdig/resource_sysdig_secure_rule_falco.go:143 +0x36d
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc00009f960, {0x10a7360, 0xc0005724e0}, 0xd?, {0xd76da0, 0xc000598690})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.0/helper/schema/resource.go:724 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0xc00009f960, {0x10a7360, 0xc0005724e0}, 0xc0006e7380, {0xd76da0, 0xc000598690})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.0/helper/schema/resource.go:1015 +0x585
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadResource(0xc0002119e0, {0x10a7360?, 0xc0005723c0?}, 0xc0006d8140)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.0/helper/schema/grpc_provider.go:613 +0x4a5
[...]

PR #227, opened 2022-12-13
fix: Solve nil pointer dereference in falco rules

tembleking:
Should solve #223


brokenpip3 (approved):

jprieto92 (approved):
LGTM, thanks @tembleking

PR #228, opened 2022-12-19
Make capture attribute required for policy resource

jacklongsd:


sysdig/resource_sysdig_secure_policy.go
```diff
@@ -98,7 +98,7 @@ func resourceSysdigSecurePolicy() *schema.Resource {
 						},
 						"capture": {
 							Type:     schema.TypeList,
-							Optional: true,
+							Required: true,
```
kmvachhani, line 101: I believe this makes the capture block required which is not what we want. we should change the optional field on line 116 to required for just the name field

ben-m-lucas (approved):

PR #229, opened 2022-12-20
feat(secure-policy): Make capture name mandatory. Fixes #197

### LINKED ISSUE #197: Missing "Capture file name" attribute in sysdig_secure_policy
When defining a policy in the UI and enabling captures, there is a "File name" field that cannot be left empty.

The resource in the provider doesn't expose that attribute, so when a policy is created via the provider, the "File name" field is left empty, which causes the capture to fail, as it is stored locally on the agent rather than on Sysdig S3.

@lindawang-sysdig @timekillerj @garyhutchins FYI

airadier:
As described in #197 , the capture name field should be mandatory.


airadier:
Duplicate from #228

PR #239, opened 2023-01-06
fix(secure-policy): Add support for name/value exceptions

ben-m-lucas:
This PR adds support for exceptions that use just name and value.

### DIFF
--- a/website/docs/r/secure_rule_falco.md
+++ b/website/docs/r/secure_rule_falco.md
@@ -81,7 +81,7 @@ For more information about the syntax of the exceptions, check the [official Fal
 Supported fields for exceptions:
 
 * `name` - (Required) The name of the exception. Only used to provide a handy name, and to potentially link together values in a later rule that has `append = true`.
-* `fields` - (Required) Contains one or more fields that will extract a value from the syscall/k8s_audit events.
+* `fields` - (Optional) Contains one or more fields that will extract a value from the syscall/k8s_audit events.
 * `comps` - (Optional) Contains comparison operators that align 1-1 with the items in the fields property.
 * `values` - (Optional) Contains tuples of values. Each item in the tuple should align 1-1 with the corresponding field
   and comparison operator. Since the value can be a string, a list of strings or a list of a list of strings, the value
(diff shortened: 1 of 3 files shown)


jacklongsd (approved):

website/docs/r/secure_rule_falco.md
```diff
@@ -78,7 +83,12 @@ Supported fields for exceptions:
...
 * `fields` - (Required) Contains one or more fields that will extract a value from the syscall/k8s_audit events.
 * `comps` - (Optional) Contains comparison operators that align 1-1 with the items in the fields property.
-* `values` - (Required) Contains tuples of values. Each item in the tuple should align 1-1 with the corresponding field and comparison operator. Since the value can be a string, a list of strings or a list of a list of strings, the value of this field must be supplied in JSON format. You can use the default `jsonencode` function to provide this value. See the usage example on the top.
+* `values` - (Optional) Contains tuples of values. Each item in the tuple should align 1-1 with the corresponding field
+  and comparison operator. Since the value can be a string, a list of strings or a list of a list of strings, the value
+  of this field must be supplied in JSON format. You can use the default `jsonencode` function to provide this value.
+  See the usage example on the top. **Required** if `fields` and `comps` are set.
+* `value` - (Optional) Contains the single value used when exception is a name/value pair. **Required** if `fields` and
+  `comps` are not set
```
kmvachhani, line 91: so if they want to define a single field exception, and they pass fields/comps, they need to use values instead of value?

ben-m-lucas: Yes, We're basically structuring it to more closely match how falco handles these rules and exceptions.
Likewise, if they're using fields and comps and it is a single value for the field and comp, it still needs to be in an array, the same way falco expects it.

kmvachhani (approved):
LGTM!

website/docs/r/secure_rule_falco.md
```diff
@@ -78,7 +83,12 @@ Supported fields for exceptions:
 * `name` - (Required) The name of the exception. Only used to provide a handy name, and to potentially link together values in a later rule that has `append = true`.
 * `fields` - (Required) Contains one or more fields that will extract a value from the syscall/k8s_audit events.
```
ombellare, line 84: Should `fields` in the document also change from Required to Optional now?

ombellare (commented):
Small question otherwise LGTM

ombellare (approved):
LGTM!

PR #248, opened 2023-02-03
feat(legacy scanning policies and assignments): Add scanning policies and assignments for legacy engine

ctolon22:
This PR adds support for scanning policy and scanning policy assignments for Legacy Engine.

Note: As scanning policy assignments (/mappings) only support GET/PUT, the destroy action returns the mapping to the default configuration (*/*:* assigned to Default policy)

### DIFF
--- a/sysdig/internal/client/secure/models.go
+++ b/sysdig/internal/client/secure/models.go
@@ -416,3 +416,37 @@ func BenchmarkTaskFromJSON(body []byte) *BenchmarkTask {
 
 	return &result
 }
+
+// -------- Scanning Policies --------
+type ScanningPolicy struct {
+	ID             string         `json:"id,omitempty"`
+	Version        string         `json:"version,omitempty"`
+	Name           string         `json:"name"`
+	Comment        string         `json:"comment"`
+	IsDefault      bool           `json:"isDefault,omitempty"`
+	PolicyBundleId string         `json:"policyBundleId,omitempty"`
+	Rules          []ScanningGate `json:"rules"`
+}
+
+type ScanningGate struct {
+	ID      string              `json:"id,omitempty"`
+	Gate    string              `json:"gate"`
+	Trigger string              `json:"trigger"`
+	Action  string              `json:"action"`
+	Params  []ScanningGateParam `json:"params"`
+}
+
+type ScanningGateParam struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (policy *ScanningPolicy) ToJSON() io.Reader {
+	payload, _ := json.Marshal(policy)
+	return bytes.NewBuffer(payload)
+}
+
+func ScanningPolicyFromJSON(body []byte) (result ScanningPolicy) {
+	_ = json.Unmarshal(body, &result)
+	return result
+}
--- a/sysdig/internal/client/secure/models.go
+++ b/sysdig/internal/client/secure/models.go
@@ -450,3 +450,34 @@ func ScanningPolicyFromJSON(body []byte) (result ScanningPolicy) {
 	_ = json.Unmarshal(body, &result)
 	return result
 }
+
+// -------- Scanning Policy Assignments --------
+type ScanningPolicyAssignmentList struct {
+	Items          []ScanningPolicyAssignment `json:"items"`
+	PolicyBundleId string                     `json:"policyBundleId"`
+}
+
+type ScanningPolicyAssignment struct {
+	ID           string                        `json:"id,omitempty"`
+	Name         string                        `json:"name"`
+	Registry     string                        `json:"registry"`
+	Repository   string                        `json:"repository"`
+	Image        ScanningPolicyAssignmentImage `json:"image"`
+	PolicyIDs    []string                      `json:"policy_ids"`
+	WhitelistIDs []string                      `json:"whitelist_ids"`
+}
+
+type ScanningPolicyAssignmentImage struct {
+	Type  string `json:"type"`
+	Value string `json:"value"`
+}
+
+func (policy *ScanningPolicyAssignmentList) ToJSON() io.Reader {
+	payload, _ := json.Marshal(policy)
+	return bytes.NewBuffer(payload)
+}
+
+func ScanningPolicyAssignmentFromJSON(body []byte) (result ScanningPolicyAssignmentList) {
+	_ = json.Unmarshal(body, &result)
+	return result
+}
--- a/sysdig/internal/client/secure/scanningpolicies.go
+++ b/sysdig/internal/client/secure/scanningpolicies.go
@@ -7,6 +7,20 @@ import (
 	"net/http"
 )
 
+func (client *sysdigSecureClient) scanningPoliciesURL() string {
+	return fmt.Sprintf("%s/api/scanning/v1/policies", client.URL)
+}
+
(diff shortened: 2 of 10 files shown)


sysdig/resource_sysdig_secure_scanningpolicies.go
```diff
@@ -0,0 +1,251 @@
...
+func resourceSysdigScanningPolicyUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := meta.(SysdigClients).sysdigSecureClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	scanningPolicy := scanningPolicyFromResourceData(d)
+	id, _ := d.Get("id").(string)
+	scanningPolicy.ID = id
+	fmt.Println(scanningPolicy)
```
ben-m-lucas, line 134: I suspect this may have been left in here for debugging purposes.  If so, we should remove this line.

```suggestion
```

jacklongsd (approved):

sysdig/resource_sysdig_secure_scanningpolicies.go
```diff
@@ -0,0 +1,251 @@
...
+						},
+						"gate": {
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: validateDiagFunc(validation.StringInSlice([]string{"always", "dockerfile", "files", "licenses", "metadata", "npms", "packages", "passwd_file", "retrieved_files", "vulnerabilities", "secret_scans", "ruby_gems"}, false)),
+						},
+						"trigger": {
+							Type:     schema.TypeString,
+							Required: true,
+							// ValidateDiagFunc: TODO: create inline func to validate each trigger options depending on gate https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors
```
ben-m-lucas, line 77: There are a couple of TODOs in the new code.  Do we need to implement these now?  Or are these reminders for a later date?

ben-m-lucas:
I just had a couple of minor comments/questions.  Otherwise, it looks good.

sysdig/resource_sysdig_secure_scanningpolicies.go
ctolon22: those are reminders of improvements that can be done in the future. At the end, the API validates the gates / policy parameters so this would be to "double-validate".

ctolon22: those are reminders of improvements that can be done in the future. At the end, the API validates the gates / policy parameters so this would be to "double-validate".

sysdig/internal/client/secure/models.go
```diff
@@ -416,3 +416,68 @@ func BenchmarkTaskFromJSON(body []byte) *BenchmarkTask {
...
+	Params  []ScanningGateParam `json:"params"`
+}
+
+type ScanningGateParam struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (policy *ScanningPolicy) ToJSON() io.Reader {
+	payload, _ := json.Marshal(policy)
```
filiptubic, line 445: Ignoring error here is bad practice but I can see that this "ignored" `ToJSON` function is copy pasted everywhere in TP code. Is there any reason why is written like this?

filiptubic: I see now that error is ignored in most of the places in code, why is that?

sysdig/internal/client/secure/scanningpolicies.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (client *sysdigSecureClient) DeleteScanningPolicyById(ctx context.Context, scanningPolicyID string) error {
+	response, err := client.doSysdigSecureRequest(ctx, http.MethodDelete, client.scanningPolicyURL(scanningPolicyID), nil)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
```
filiptubic, line 87: Do we have this condition here because we do not know what is the 2xx respond code of the DELETE action or server could return either of those two http codes?

sysdig/resource_sysdig_secure_scanningpolicies.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func resourceSysdigScanningPolicyUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := meta.(SysdigClients).sysdigSecureClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	scanningPolicy := scanningPolicyFromResourceData(d)
+	id, _ := d.Get("id").(string)
```
filiptubic, line 131: Shouldn't we get error here and check it? If its error return:
```
return diag.FromErr(err)
```
wdyt?

sysdig/resource_sysdig_secure_scanningpolicies_test.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		},
+		ProviderFactories: map[string]func() (*schema.Provider, error){
+			"sysdig": func() (*schema.Provider, error) {
+				return sysdig.Provider(), nil
+			},
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: scanningPolicyWithName(rText()),
```
filiptubic, line 30: Im sorry if this does not makes sense (since Im new with TP) but should we add some validation here?
Based on official docs there is a way to check created resource by using some of the builtin check funcs: 
https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/teststep#builtin-check-functions

filiptubic: No blocker here, this is more for me the understand why we do not perform any checks here.

sysdig/internal/client/secure/models.go
ctolon22: afaik, terraform validates the JSON is "valid" in format and later we do the request where we grab the error code. (i am new to terraform development so we can add error validation there if it's necessary)

sysdig/internal/client/secure/scanningpolicies.go
ctolon22: Per this [test](https://github.com/draios/secure-backend/blob/main/scanning/legacy/backend/test/api/integration/test/policies_test.go#L83) for that same endpoint in our backend we check the `StatusNoContent`

But, checking [backend code](https://github.com/draios/secure-backend/blob/main/scanning/legacy/backend/api/api.go#L1212) seems we send the `StatusNoContent` as well as result

sysdig/resource_sysdig_secure_scanningpolicies.go
ctolon22: uhm, that `d.Get("id")` is for the resource that previously has been created.
You run create, that creates in the state the resource with the ID and the run the update/read/delete. 
Checking further, don't think the `Get` returns error so the `_` should be removed. I took as a reference others so that's why is the same than other resources. 
Removing those for this PR.

filiptubic (approved):
LGTM 🚀

kmvachhani (approved):

PR #251, opened 2023-02-08
feat(monitor,secure): support different regions for opsgenie notification channel

dbonf:
Add option to specify the region of the OpsGenie service.

### DIFF
--- a/sysdig/resource_sysdig_monitor_notification_channel_opsgenie_test.go
+++ b/sysdig/resource_sysdig_monitor_notification_channel_opsgenie_test.go
@@ -38,6 +38,14 @@ func TestAccMonitorNotificationChannelOpsGenie(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
+			{
+				Config: monitorNotificationChannelOpsGenieWithNameAndRegion(rText()),
+			},
+			{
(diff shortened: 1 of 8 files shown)


sysdig/resource_sysdig_monitor_notification_channel_opsgenie_test.go
```diff
@@ -52,3 +60,15 @@ resource "sysdig_monitor_notification_channel_opsgenie" "sample-opsgenie" {
 	notify_when_resolved = false
 }`, name)
 }
+
+func monitorNotificationChannelOpsGenieWithNameAndRegion(name string) string {
```
wideawakening, line 64: we need to update the secure brother too
https://github.com/sysdiglabs/terraform-provider-sysdig/blob/master/sysdig/resource_sysdig_secure_notification_channel_opsgenie_test.go

wideawakening (approved):

sysdig/resource_sysdig_monitor_notification_channel_opsgenie_test.go
dbonf: you are absolutely right, I added the same test also there now.

hayk99 (approved):

PR #265, opened 2023-03-09
Fix Issue With Handling Missing Policy

jacklongsd:

### DIFF
--- a/sysdig/internal/client/secure/client.go
+++ b/sysdig/internal/client/secure/client.go
@@ -15,7 +15,7 @@ type SysdigSecureClient interface {
 	CreatePolicy(context.Context, Policy) (Policy, error)
 	DeletePolicy(context.Context, int) error
 	UpdatePolicy(context.Context, Policy) (Policy, error)
-	GetPolicyById(context.Context, int) (Policy, error)
+	GetPolicyById(context.Context, int) (Policy, error, int)
 
 	CreateRule(context.Context, Rule) (Rule, error)
 	GetRuleByID(context.Context, int) (Rule, error)
--- a/sysdig/internal/client/secure/client.go
+++ b/sysdig/internal/client/secure/client.go
@@ -15,7 +15,7 @@ type SysdigSecureClient interface {
 	CreatePolicy(context.Context, Policy) (Policy, error)
 	DeletePolicy(context.Context, int) error
 	UpdatePolicy(context.Context, Policy) (Policy, error)
-	GetPolicyById(context.Context, int) (Policy, error, int)
+	GetPolicyById(context.Context, int) (Policy, int, error)
 
 	CreateRule(context.Context, Rule) (Rule, error)
 	GetRuleByID(context.Context, int) (Rule, error)
--- a/sysdig/internal/client/secure/policies.go
+++ b/sysdig/internal/client/secure/policies.go
@@ -67,7 +67,7 @@ func (client *sysdigSecureClient) UpdatePolicy(ctx context.Context, policyReques
 	return PolicyFromJSON(body), nil
 }
 
-func (client *sysdigSecureClient) GetPolicyById(ctx context.Context, policyID int) (policy Policy, err error, statusCode int) {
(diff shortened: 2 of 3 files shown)


sysdig/resource_sysdig_secure_policy.go
```diff
@@ -258,7 +259,9 @@ func resourceSysdigPolicyRead(ctx context.Context, d *schema.ResourceData, meta
 
 	if err != nil {
 		d.SetId("")
-		return diag.FromErr(err)
+		if !strings.Contains(err.Error(), policyNotFound) {
```
ombellare, line 262: Is it possible to use the status code 404 rather than the error message? The only concern is that someone might not know that the error message is being used by the terraform provider and unintentionally breaks it

jacklongsd: The way it is setup, the status code is not available at this level. And I am worried about making changes beyond this area due to lack of understanding about how it would affect other components.

sysdig/internal/client/secure/policies.go
```diff
@@ -67,20 +67,20 @@ func (client *sysdigSecureClient) UpdatePolicy(ctx context.Context, policyReques
 	return PolicyFromJSON(body), nil
 }
 
-func (client *sysdigSecureClient) GetPolicyById(ctx context.Context, policyID int) (policy Policy, err error) {
+func (client *sysdigSecureClient) GetPolicyById(ctx context.Context, policyID int) (policy Policy, err error, statusCode int) {
```
ben-m-lucas, line 70:
```suggestion
func (client *sysdigSecureClient) GetPolicyById(ctx context.Context, policyID int) (policy Policy, statusCode int, err error) {
```

In idiomatic go, the convention is to have the error return value as the last value returned.

sysdig/internal/client/secure/policies.go
```diff
@@ -67,20 +67,20 @@ func (client *sysdigSecureClient) UpdatePolicy(ctx context.Context, policyReques
...
-		return Policy{}, errorFromResponse(response)
+		return Policy{}, errorFromResponse(response), response.StatusCode
 	}
 
 	body, err := io.ReadAll(response.Body)
 	if err != nil {
 		return
 	}
-	return PolicyFromJSON(body), nil
+	return PolicyFromJSON(body), nil, http.StatusOK
```
ben-m-lucas, line 85:
```suggestion
	return PolicyFromJSON(body), http.StatusOK, nil
```

sysdig/resource_sysdig_secure_policy.go
```diff
@@ -248,18 +249,17 @@ func addActionsToPolicy(d *schema.ResourceData, policy *secure.Policy) {
...
 func resourceSysdigPolicyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	const policyNotFound = "failed to find a policy with that ID"
 	client, err := meta.(SysdigClients).sysdigSecureClient()
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
 	id, _ := strconv.Atoi(d.Id())
-	policy, err := client.GetPolicyById(ctx, id)
+	policy, err, statusCode := client.GetPolicyById(ctx, id)
```
ben-m-lucas, line 258:
```suggestion
	policy, statusCode, err := client.GetPolicyById(ctx, id)
```

sysdig/internal/client/secure/client.go
```diff
@@ -15,7 +15,7 @@ type SysdigSecureClient interface {
 	CreatePolicy(context.Context, Policy) (Policy, error)
 	DeletePolicy(context.Context, int) error
 	UpdatePolicy(context.Context, Policy) (Policy, error)
-	GetPolicyById(context.Context, int) (Policy, error)
+	GetPolicyById(context.Context, int) (Policy, error, int)
```
ben-m-lucas, line 18:
```suggestion
	GetPolicyById(context.Context, int) (Policy, int, error)
```

ben-m-lucas (approved):

kmvachhani (approved):

ombellare (approved):
LGTM. Thanks for addressing the concern!

PR #268, opened 2023-03-17
fix(falco): json omitempty on description and output

filiptubic:
Omit empty description and output when making falco request.


ben-m-lucas (approved):

sysdig/internal/client/secure/models.go
```diff
@@ -125,7 +125,7 @@ type notificationChannelWrapper struct {
 type Rule struct {
 	ID          int      `json:"id,omitempty"`
 	Name        string   `json:"name"`
```
ombellare, line 127: Should name also add the omitempty tag as it is also not allowed to be empty?

ombellare: I think this might not be needed as backend needs to throw an error on empty rule name, so sending "" should be fine.

PR #278, opened 2023-03-23
fix: Remove tee comand from output as it was overrinding the final result

AlvarVG:
Remove tee comand from output as it was overrinding the final result

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -34,11 +34,11 @@ test: fmtcheck
 	go test $(TEST) -timeout=30s -parallel=4
 
 testacc: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race 2>&1 | tee testacc-output.txt
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race
 
-junit-report: testacc
+junit-report: fmtcheck
 	@go install github.com/jstemmer/go-junit-report/v2@latest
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -34,11 +34,11 @@ test: fmtcheck
...
 
 testacc: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race 2>&1 | tee testacc-output.txt
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race
 
-junit-report: testacc
+junit-report: fmtcheck
 	@go install github.com/jstemmer/go-junit-report/v2@latest
-	@go-junit-report -in testacc-output.txt -out junit-report.xml
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race 2>&1 | go-junit-report -iocopy -out junit-report.xml
```
filiptubic, line 41: This will not work either. Piping the command will result in shadowing previous exit code from command
```
CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -race 2>&1
```

AlvarVG: It does work for us, because we are not waiting for the command exit code. We wait for the junit report which included all the failures and passes with each stdout. 

That is why i duplicated the code. To do not interfere with the normal workflow you have.

filiptubic (approved):

PR #299, opened 2023-04-21
fix(serverless-agent): mark the access key as optional

francesco-racciatti:
# Scope
Serverless Agent

# Description
This PR marks the access key as optional in the fargate workload agent data source.
Note that:
 - the direct connection mode has been deprecated,
 - and the workload agent does not require the access key to connect to the orchestrator agent.

### DIFF
--- a/website/docs/d/fargate_workload_agent.md
+++ b/website/docs/d/fargate_workload_agent.md
@@ -22,8 +22,7 @@ data "sysdig_fargate_workload_agent" "instrumented_containers" {
 
   image_auth_secret    = ""
   workload_agent_image = "quay.io/sysdig/workload-agent:latest"
-
-  sysdig_access_key = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+  
   orchestrator_host    = module.fargate-orchestrator-agent.orchestrator_host
   orchestrator_port    = module.fargate-orchestrator-agent.orchestrator_port
 }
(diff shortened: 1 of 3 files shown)


website/docs/d/fargate_workload_agent.md
```diff
@@ -32,7 +31,6 @@ data "sysdig_fargate_workload_agent" "instrumented_containers" {
 ## Argument Reference
 
 * `container_definitions` - (Required) The input Fargate container definitions to instrument with the Sysdig workload agent.
-* `sysdig_access_key` - (Required) The Sysdig Access Key (Agent token).
 * `orchestrator_host` - (Required) The orchestrator host to connect to.
 * `orchestrator_port` - (Required) The orchestrator port to connect to.
```
francesco-racciatti, line 35: Just a heads up

These arguments are marked as required, but currently, that's not really enforced.

website/docs/d/fargate_workload_agent.md
```diff
@@ -32,7 +31,6 @@ data "sysdig_fargate_workload_agent" "instrumented_containers" {
 ## Argument Reference
 
 * `container_definitions` - (Required) The input Fargate container definitions to instrument with the Sysdig workload agent.
-* `sysdig_access_key` - (Required) The Sysdig Access Key (Agent token).
```
pgcrooks-sysdig, line 35: I think we should keep this in the docs as "optional" since we do still consume it. We may have to resurrect this feature in the future.
This isn't a mandatory ask from me, happy either way.

francesco-racciatti: I prefer removing it from the doc, at least for now, as we did for the collector host/port a while ago.
There's no point in passing the access key to the workload agent ATM unless you want to use the direct connection mode.

pgcrooks-sysdig: Fair enough, happy to remove it.

PR #300, opened 2023-04-21
test: Add build labels to properly group monitor resources as well as ibm ones

AlvarVG:
* Adding proper labels to the monitor resources
* Marking ibm only resources with label tf_acc_ibm
* Setting parallelization over testacc

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -38,7 +38,7 @@ testacc: fmtcheck
 	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race -parallel=4
 
 testacc-ibm: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_ibm -timeout 120m -race -parallel=4
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_monitor,tf_acc_ibm -timeout 120m -race -parallel=4
 
 junit-report: fmtcheck
 	@go install github.com/jstemmer/go-junit-report/v2@latest
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -35,14 +35,14 @@ test: fmtcheck
 	go test $(TEST) -tags=unit -timeout=30s -parallel=4
 
 testacc: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race -parallel=1
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race -parallel=4
 
 testacc-ibm: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_ibm -timeout 120m -race -parallel=1
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_ibm -timeout 120m -race -parallel=4
 
 junit-report: fmtcheck
 	@go install github.com/jstemmer/go-junit-report/v2@latest
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race 2>&1 -parallel=1 | go-junit-report -iocopy -out junit-report.xml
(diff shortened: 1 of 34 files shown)


GNUmakefile
```diff
@@ -35,14 +35,14 @@ test: fmtcheck
...
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race -parallel=4
 
 testacc-ibm: fmtcheck
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_ibm -timeout 120m -race -parallel=1
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=tf_acc_monitor,tf_acc_ibm -timeout 120m -race -parallel=4
 
 junit-report: fmtcheck
 	@go install github.com/jstemmer/go-junit-report/v2@latest
-	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race 2>&1 -parallel=1 | go-junit-report -iocopy -out junit-report.xml
+	CGO_ENABLED=1 TF_ACC=1 go test $(TEST) -v $(TESTARGS) -tags=$(TEST_SUITE) -timeout 120m -race 2>&1 -parallel=4 | go-junit-report -iocopy -out junit-report.xml
```
papajulio, line 45: Nope sorry we can not put this to 4 here as it adds a lot of instability to the secure policies tests, if we want we can make this configurable and you can customize that in your jobs but for them to be stable and pass in general better have it to 1 I would say.

filiptubic: I agree with @papajulio

sysdig/data_source_sysdig_current_user_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build tf_acc_sysdig
+//go:build tf_acc_sysdig || tf_acc_monitor
```
filiptubic, line 2: By specifying `tf_acc_monitor` you are going to run test on IBM which you should not.

filiptubic: Very few resources are refactored to use new v2 client. And only they have ability to run on IBM. User resources/data sources right now do not have this feature.

sysdig/data_source_sysdig_monitor_notification_channel_email_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build tf_acc_sysdig || tf_acc_ibm
+//go:build tf_acc_sysdig || tf_acc_monitor
```
filiptubic, line 1: this will have the same effect

sysdig/data_source_sysdig_user_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build tf_acc_sysdig
+//go:build tf_acc_sysdig || tf_acc_monitor
```
filiptubic, line 1: This will always fail on IBM, as said in previous comments

sysdig/resource_sysdig_monitor_alert_anomaly_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build tf_acc_sysdig
+//go:build tf_acc_sysdig || tf_acc_monitor
```
filiptubic, line 1: alerts are not using v2 client, again wont work

filiptubic (changes requested):
@AlvarVG lets have a meeting about your PR, and let me give you a quick introduction how we are running tests in terraform. If I understood your intentions, you wanted to mark tests as monitor and that will allow you to run only monitor tests regardless if its sysdig or IBM environment. That is not possible with current github actions. As you can see your tests are failing with missing environment variables. One more note is that you expressions in //go:build comments are wrong, eg
```
//go:build tf_acc_sysdig || tf_acc_monitor
```
this will be executed if any of those two tags are provided, which means having `-tags=tf_acc_monitor,tf_acc_ibm` will execute test with above statement.

The only change I would approve at this moment is renaming `tf_acc_ibm` to `tf_acc_ibm_monitor` because we are (for now) testing only monitor on IBM and this renaming will happen eventually once we add integration with IBM Cloud Secure.

GNUmakefile
AlvarVG: Tested for monitor resources and it is working fine...

sysdig/resource_sysdig_monitor_alert_anomaly_test.go
AlvarVG: v1 is still alive and we are using it for multiple scenarios in our framework, so there is no problem at all.

filiptubic: V1 is still alive but terraform provider does not support it. You cant use v1 alerts resources via terraform ATM.

PR #303, opened 2023-04-25
refactor(tests): add more granular test tags (per env per product)

filiptubic:
Here we want to add more granularity in our testing. We want to add ability to run either monitor or secure test suite on Sysdig or IBM environment. To do that we added following build tags:
- `tf_acc_sysdig_monitor`
- `tf_acc_sysdig_secure`
- `tf_acc_ibm_monitor`

Github actions are refactored as well to run tests per product in parallel.

### DIFF
--- a/sysdig/resource_sysdig_user_test.go
+++ b/sysdig/resource_sysdig_user_test.go
@@ -4,7 +4,6 @@ package sysdig_test
 
 import (
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
@@ -16,7 +15,6 @@ import (
 
 func TestAccUser(t *testing.T) {
 	rText := func() string { return acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum) }
-	customdiff.All()
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck: preCheckAnyEnv(t, SysdigMonitorApiTokenEnv, SysdigSecureApiTokenEnv),
 		ProviderFactories: map[string]func() (*schema.Provider, error){
--- a/sysdig/resource_sysdig_user_test.go
+++ b/sysdig/resource_sysdig_user_test.go
@@ -4,6 +4,7 @@ package sysdig_test
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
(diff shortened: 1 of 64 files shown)


sysdig/resource_sysdig_user_test.go
```diff
@@ -48,35 +50,35 @@ func TestAccUser(t *testing.T) {
...
 func userWithoutSystemRole(name string) string {
 	return fmt.Sprintf(`
 resource "sysdig_user" "sample" {
-  email      = "terraform-test+user@sysdig.com"
-  first_name = "%s"
-  last_name  = "%s"
+  email      = "terraform-test+user%[1]s@sysdig.com"
+  first_name = "%[1]s"
+  last_name  = "%[1]s"
 }`, name, name)
```
francesco-racciatti, line 76: The 3rd argument of `Sprintf` (i.e. the 2nd `name`) looks not necessary since we are using `%[1]s` always

sysdig/resource_sysdig_user_test.go
```diff
@@ -48,35 +50,35 @@ func TestAccUser(t *testing.T) {
...
 }
 
-func userMinimumConfiguration() string {
-	return `
+func userMinimumConfiguration(name string) string {
+	return fmt.Sprintf(`
 resource "sysdig_user" "sample" {
-  email      = "terraform-test+user@sysdig.com"
-}`
+  email      = "terraform-test+user%[1]s@sysdig.com"
```
francesco-racciatti, line 82: `%s` should be sufficient here

sysdig/resource_sysdig_user_test.go
```diff
@@ -25,16 +27,16 @@ func TestAccUser(t *testing.T) {
...
 			},
 			{
-				Config: userWithName(rText()),
+				Config: userWithName(name),
 			},
 			{
-				Config: userWithoutSystemRole(rText()),
+				Config: userWithoutSystemRole(name),
 			},
 			{
```
francesco-racciatti, line 38: QQ Previously, each configuration had a different random name, but now all configurations take the same random name. Is this the intended behavior?

sysdig/resource_sysdig_user_test.go
filiptubic: good catch

sysdig/resource_sysdig_user_test.go
filiptubic: I was debugging some tests and thought it was the issue. In the end issue was that the same email `terraform-test+user@sysdig.com` was used for all tests. So, since we now have secure tests running in parallel they try to create user with the same email and fail. This is some leftover and tbh it wont affect tests at all but I will revert those changes just to be sure.

francesco-racciatti (approved):
LGTM

sysdig/resource_sysdig_user_test.go
filiptubic: As I said previously `terraform-test+user@sysdig.com` was the same email used across all tests. So resource `sysdig_user.sample` cant change username (which is email) overtime. This is the reason why tests fails with infra drifts. Now I remember why I had this change 😄

PR #308, opened 2023-05-03
Default state in a policy actions block should not result in unexpected state change in terraform

ben-m-lucas:
When the actions block in a module is empty, this can result in a condition in which terraform is identifying the resource as having changed.  This PR changes this so it identifies the default state.

### DIFF
--- a/sysdig/resource_sysdig_secure_policy.go
+++ b/sysdig/resource_sysdig_secure_policy.go
@@ -176,7 +176,14 @@ func policyToResourceData(policy *secure.Policy, d *schema.ResourceData) {
 			}}
 		}
 	}
-	_ = d.Set("actions", actions)
+
+	currentContainerAction := d.Get("actions.0.container").(string)
+	currentCaptureAction := d.Get("actions.0.capture").([]interface{})
+	// If the policy retrieved from service has no actions and the current state is default values,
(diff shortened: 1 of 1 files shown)


sysdig/resource_sysdig_secure_policy.go
```diff
@@ -176,7 +176,14 @@ func policyToResourceData(policy *secure.Policy, d *schema.ResourceData) {
 			}}
 		}
 	}
-	_ = d.Set("actions", actions)
+
+	currentContainerAction := d.Get("actions.0.container").(string)
+	currentCaptureAction := d.Get("actions.0.capture").([]interface{})
+	// If the policy retrieved from service has no actions and the current state is default values,
+	// then do not set the "actions" key as it may cause terraform to think there has been a state change
+	if len(policy.Actions) > 0 || currentContainerAction != "" || len(currentCaptureAction) > 0 {
```
ombellare, line 184: Would this also work when policy has other action types - pause, stop, prevent drift, kill associated with it?

ben-m-lucas: Yes, the only time `len(policy.Actions)` is 0 is when there is no action associated with it.  Likewise, if it has one of the other action types, `currentContainerAction` will not be ""

ombellare (approved):

PR #309, opened 2023-05-04
refactor(notificationchannels): refactor secure notification channels to use v2 client

filiptubic:
Here we refactored notification channels (resources and data sources) to use V2 client.
In addition all unused code from secure models are deleted.


draraksysdig (approved):
LGTM

PR #312, opened 2023-05-04
refactor(policies): refactor policies to use v2 client

filiptubic:
Here we refactored policies resource on secure to use v2 client.


draraksysdig (approved):
LGTM

PR #316, opened 2023-05-05
feat(policies): add resource for managing the configuration of secure managed policies

ben-m-lucas:
This PR adds a new resource type `sysdig_secure_managed_policy` that can be used for managed policies to:
* enable/disable the pollcy
* enable/disable rules within the policy
* define the scope, runbook and action for the policy


sysdig/resource_sysdig_secure_managed_policy.go
```diff
@@ -0,0 +1,307 @@
...
+			},
+			"runbook": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"actions": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
ombellare, line 81: As this action object would be present in managed ruleset and custom policies as well, does it make sense to abstract it into model.go?

sysdig/resource_sysdig_secure_managed_policy.go
```diff
@@ -0,0 +1,307 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"type": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "falco",
+				ValidateDiagFunc: validateDiagFunc(validation.StringInSlice([]string{"falco", "list_matching", "k8s_audit", "aws_cloudtrail", "gcp_auditlog", "azure_platformlogs"}, false)),
```
ombellare, line 43: Does it make sense to abstract this validateDiagFunc into model.go as well as it will be used for all policy type resources?

ombellare (commented):
Couple of small optional suggestions. Please let me know if you think they make sense. If needed those can be done in a separate PR and I can approve this

sysdig/resource_sysdig_secure_managed_policy.go
ben-m-lucas: That's a great point.  We also have more policy resource types coming so this could really help out there.  I'll take a look.

kmvachhani (approved):
would agree with Omar's suggestions but otherwise LGTM

PR #330, opened 2023-05-12
fix(tests): add tf_acc_policies tag in common_test

filiptubic:
This should fix missing common test utils when running tf_acc_policies test suite.


sysdig/common_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build tf_acc_sysdig || tf_acc_sysdig_monitor || tf_acc_sysdig_secure || tf_acc_ibm || tf_acc_ibm_monitor || unit
+//go:build tf_acc_sysdig || tf_acc_sysdig_monitor || tf_acc_sysdig_secure || tf_acc_ibm || tf_acc_ibm_monitor || tf_acc_policies || unit
```
jcelaya, line 1: Isn't there an option to include this always? I know Secure folks will want to use more tags in the future (and probably us in Monitor too) cc @papajulio

filiptubic: Im not sure if we can automagically add those build tags. Its getting bigger and bigger we might need to investigate other options how to make it easier for maintenance. For now this will fix the problem.

jcelaya (approved):

sysdig/common_test.go
papajulio: Probably the best way that I can think of to achieve that is to define a common label and always pass it from the Makefile in all the definitions we have there I would say, I don't know if there is another way to achieve that, so we avoid to increase the number of labels here in this file....

filiptubic: One of the options is to use `ldflags` to mark which test suite you are going to run (in makefile). After seeing this sausage of pipes I more and more like this idea 😄

draraksysdig (approved):

PR #349, opened 2023-05-19
Remove unused tags

filiptubic:
- Removed unused tags `tf_acc_sysdig` ,`tf_acc_ibm`
- Lint is part of test workflow, removed code duplication
- `golangci-lint` will lint tests which contains build tags

### DIFF
--- a/sysdig/common_test.go
+++ b/sysdig/common_test.go
@@ -1,5 +1,3 @@
-//go:build tf_acc_sysdig || tf_acc_sysdig_monitor || tf_acc_sysdig_secure || tf_acc_ibm || tf_acc_ibm_monitor || tf_acc_policies || unit
-
 package sysdig_test
 
 import (
--- a/sysdig/internal/client/v2/ibm_test.go
+++ b/sysdig/internal/client/v2/ibm_test.go
@@ -168,7 +168,6 @@ func TestIBMClient_CurrentTeamID(t *testing.T) {
 		if err != nil {
 			t.Errorf("failed to create response, err: %v", err)
 		}
-		return
 	}))
 
 	for _, testCase := range testTable {
--- a/sysdig/resource_sysdig_secure_managed_policy_test.go
+++ b/sysdig/resource_sysdig_secure_managed_policy_test.go
@@ -91,7 +91,7 @@ resource "sysdig_secure_managed_policy" "sample" {
 }
 
 func managedPolicyWithoutNotificationChannels() string {
-	return fmt.Sprintf(`
+	return `
 resource "sysdig_secure_managed_policy" "sample" {
 	name = "Sysdig Runtime Threat Detection"
 	enabled = true
@@ -107,22 +107,19 @@ resource "sysdig_secure_managed_policy" "sample" {
 		  name = "testcapture"
 		}
 	}	
-}
-	`)
+}`
 }
 
 func managedPolicyWithMinimumConfiguration() string {
-	return fmt.Sprintf(`
+	return `
 resource "sysdig_secure_managed_policy" "sample" {
 	name = "Sysdig Runtime Threat Detection"
 	enabled = true
-}
-	`)
+}`
 }
 
 func managedPolicyWithKillAction() string {
-	return fmt.Sprintf(`
-resource "sysdig_secure_managed_policy" "sample" {
+	return `resource "sysdig_secure_managed_policy" "sample" {
 	name = "Sysdig Runtime Threat Detection"
 	enabled = true
 	scope = "container.id != \"\""
@@ -132,6 +129,5 @@ resource "sysdig_secure_managed_policy" "sample" {
 	actions {
 		container = "kill"
 	}
-}
-	`)
+}`
 }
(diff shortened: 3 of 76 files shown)


sysdig/common_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build tf_acc_sysdig || tf_acc_sysdig_monitor || tf_acc_sysdig_secure || tf_acc_ibm || tf_acc_ibm_monitor || tf_acc_policies || unit
-
 package sysdig_test
```
filiptubic, line 1: Completely removing build tags will result in always building this file which we want.

francesco-racciatti: I'm not familiar with the build & test workflow of the provider, so take this comment with a grain of salt...
The side effect of completely removing this build directive will be that `common_test` will be included in any build, including release builds.

A different approach could be to set a build directive like
```
//go:build !release
```

sysdig/resource_sysdig_secure_managed_policy_test.go
```diff
@@ -107,22 +107,19 @@ resource "sysdig_secure_managed_policy" "sample" {
 		  name = "testcapture"
 		}
 	}	
-}
-	`)
```
francesco-racciatti, line 111: Probably this does not matter, but the returned string changed

sysdig/resource_sysdig_secure_managed_policy_test.go
```diff
@@ -107,22 +107,19 @@ resource "sysdig_secure_managed_policy" "sample" {
...
 }
 
 func managedPolicyWithMinimumConfiguration() string {
-	return fmt.Sprintf(`
+	return `
 resource "sysdig_secure_managed_policy" "sample" {
 	name = "Sysdig Runtime Threat Detection"
 	enabled = true
-}
-	`)
```
francesco-racciatti, line 120: The same here

sysdig/resource_sysdig_secure_managed_ruleset_test.go
```diff
@@ -127,12 +127,11 @@ resource "sysdig_secure_managed_ruleset" "sample" {
 		  name = "testcapture"
 		}
 	}	
-}
-	`)
```
francesco-racciatti, line 131: And here

sysdig/resource_sysdig_secure_managed_ruleset_test.go
```diff
@@ -162,6 +160,5 @@ resource "sysdig_secure_managed_ruleset" "sample" {
 	actions {
 		container = "kill"
 	}
-}
-	`)
```
francesco-racciatti, line 166: And here

sysdig/resource_sysdig_secure_managed_policy_test.go
```diff
@@ -107,22 +107,19 @@ resource "sysdig_secure_managed_policy" "sample" {
...
+}`
 }
 
 func managedPolicyWithKillAction() string {
-	return fmt.Sprintf(`
```
francesco-racciatti, line 124: And here

sysdig/internal/client/v2/ibm_test.go
```diff
@@ -168,7 +168,6 @@ func TestIBMClient_CurrentTeamID(t *testing.T) {
 		if err != nil {
 			t.Errorf("failed to create response, err: %v", err)
 		}
-		return
```
francesco-racciatti, line 171: This is out of the scope of this PR

francesco-racciatti (changes requested):
Looks good, but it might be worth opening a new PR for the out-of-scape changes

sysdig/common_test.go
filiptubic: AFIK go build command excludes all *_.test.go files so test files will never be included in final binary. There is no problem here actually.

filiptubic: Tags in tests are used to differentiate tests by nature, there are no connection with actual release.

sysdig/resource_sysdig_secure_managed_policy_test.go
filiptubic: Thats ok because `fmt.Sprintf` is removed since linter is failing after modifying `golangci-lint` command. Check here: https://github.com/sysdiglabs/terraform-provider-sysdig/pull/349/files/9d3c9ba36d3d4e691327a43d5c2b102c4d100ef9#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R18

francesco-racciatti: Yup. Thanks!

sysdig/internal/client/v2/ibm_test.go
filiptubic: Its not, actually this is changed due to linting stuff. I had to change it.

filiptubic: Why its not failing before? Because linter was ignoring `*_test.go` files with tags.

francesco-racciatti (approved):
LGTM

PR #363, opened 2023-06-01
feat(group mapping): added new `group_mapping_config` resource

Shadow649:
Changes made in the PR:
This PR adds support for managing group mapping conflicts resolution configuration ([doc](https://docs.sysdig.com/en/docs/administration/administration-settings/user-and-team-administration/group-mappings/#group-mapping-behavior)): 
* created new resource `sysdig_group_mapping_config` this configuration is global per customer
* acceptance tests
* documentation

### DIFF
--- a/sysdig/internal/client/v2/group_mapping_config.go
+++ b/sysdig/internal/client/v2/group_mapping_config.go
@@ -1,0 +1,105 @@
+package v2
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+var GroupMappingConfigNotFound = errors.New("group mapping configuration not found")
+
+const (
+	CreateGroupMappingConfigPath = "%s/api/groupmappings/settings"
+	UpdateGroupMappingConfigPath = "%s/api/groupmappings/settings"
+	GetGroupMappingConfigPath    = "%s/api/groupmappings/settings"
+)
+
+type GroupMappingConfigInterface interface {
+	Base
+	CreateGroupMappingConfig(ctx context.Context, gmc *GroupMappingConfig) (*GroupMappingConfig, error)
+	UpdateGroupMappingConfig(ctx context.Context, gmc *GroupMappingConfig) (*GroupMappingConfig, error)
+	GetGroupMappingConfig(ctx context.Context) (*GroupMappingConfig, error)
+}
+
+func (client *Client) CreateGroupMappingConfig(ctx context.Context, gmc *GroupMappingConfig) (*GroupMappingConfig, error) {
+	payload, err := Marshal(gmc)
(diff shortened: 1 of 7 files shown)


sysdig/resource_sysdig_group_mapping_config_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"os"
+	"testing"
+)
+
+func TestAccGroupMappingConfig(t *testing.T) {
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
```
filiptubic, line 16: Change this PreCheck similar to this: https://github.com/sysdiglabs/terraform-provider-sysdig/blob/master/sysdig/resource_sysdig_monitor_alert_downtime_test.go#L20

We do not want to copy-paste this same block of code all over the place. This function can help us :)

sysdig/resource_sysdig_group_mapping_config.go
```diff
@@ -0,0 +1,120 @@
...
+func groupMappingConfigToResourceData(groupMappingConfig *v2.GroupMappingConfig, d *schema.ResourceData) error {
+	err := d.Set("no_mapping_strategy", groupMappingConfig.NoMappingStrategy)
+	if err != nil {
+		return err
+	}
+	err = d.Set("different_team_same_role_strategy", groupMappingConfig.DifferentTeamSameRoleStrategy)
+	if err != nil {
+		return err
+	}
+	return nil
```
filiptubic, line 112: you could just use 
```
return d.Set("different_team_same_role_strategy", groupMappingConfig.DifferentTeamSameRoleStrategy)
```

sysdig/internal/client/v2/group_mapping_config.go
```diff
@@ -0,0 +1,105 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+var GroupMappingConfigNotFound = errors.New("group mapping configuration not found")
+
+const (
+	CreateGroupMappingConfigPath = "%s/api/groupmappings/settings"
```
filiptubic, line 13: We have here only Get and Put requests on the same URL path. Lets just have one variable `GroupMappingConfigPath="%s/api/groupmappings/settings"`

I would also remove `CreateGroupMappingConfig` since its doing exactly the same thing as update. I would just keep `GetGroupMappingConfig` and `UpdateGroupMappingConfig`

filiptubic (approved):

PR #368, opened 2023-06-09
Support new Secure notification templates for Slack/Teams

ombellare:


sysdig/resource_sysdig_secure_notification_channel_msteams.go
```diff
@@ -0,0 +1,213 @@
...
+	client, err := getSecureNotificationChannelClient(meta.(SysdigClients))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, _ := strconv.Atoi(d.Id())
+	nc, err := client.GetNotificationChannelById(ctx, id)
+
+	if err != nil {
+		d.SetId("")
```
filiptubic, line 85: So if there is internal server error we are removing resource from state? Why are we removing resource from state when there is an error from API?

sysdig/resource_sysdig_secure_notification_channel_msteams.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func resourceSysdigSecureNotificationChannelMSTeamsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := getSecureNotificationChannelClient(meta.(SysdigClients))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, _ := strconv.Atoi(d.Id())
```
filiptubic, line 81: please do not ignore errors, here and also in other places

sysdig/resource_sysdig_secure_notification_channel_msteams.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	notificationChannel, err = client.CreateNotificationChannel(ctx, notificationChannel)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.Itoa(notificationChannel.ID))
```
filiptubic, line 69: Instaed of doing this d.Set call `resourceSysdigSecureNotificationChannelMSTeamsRead` which will populate everything.

sysdig/resource_sysdig_secure_notification_channel_msteams.go
```diff
@@ -0,0 +1,213 @@
...
+		return diag.FromErr(err)
+	}
+
+	nc, err := secureNotificationChannelMSTeamsFromResourceData(d, teamID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	nc.Version = d.Get("version").(int)
+	nc.ID, _ = strconv.Atoi(d.Id())
```
filiptubic, line 113: ignored err

sysdig/resource_sysdig_secure_notification_channel_msteams.go
```diff
@@ -0,0 +1,213 @@
...
+
+	_, err = client.UpdateNotificationChannel(ctx, nc)
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
filiptubic, line 118: call `resourceSysdigSecureNotificationChannelMSTeamsRead` after update.

sysdig/resource_sysdig_secure_notification_channel_msteams_test.go
```diff
@@ -0,0 +1,73 @@
...
+			"sysdig": func() (*schema.Provider, error) {
+				return sysdig.Provider(), nil
+			},
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: secureNotificationChannelMSTeamsWithName(rText()),
+			},
+			{
+				ResourceName:      "sysdig_secure_notification_channel_msteams.sample-msteams",
```
filiptubic, line 32: Maybe test import after all other steps done. Usually we test all updates then import

sysdig/provider.go
```diff
@@ -114,6 +114,7 @@ func Provider() *schema.Provider {
 			"sysdig_secure_notification_channel_webhook":   resourceSysdigSecureNotificationChannelWebhook(),
 			"sysdig_secure_notification_channel_slack":     resourceSysdigSecureNotificationChannelSlack(),
 			"sysdig_secure_notification_channel_pagerduty": resourceSysdigSecureNotificationChannelPagerduty(),
+			"sysdig_secure_notification_channel_msteams":   resourceSysdigSecureNotificationChannelMSTeams(),
```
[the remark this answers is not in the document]
ombellare: Makes sense, let me add the same to monitor as well 👍

sysdig/resource_sysdig_secure_notification_channel_msteams.go
ombellare: I used the logic from another file, but I think the logic is to remove state from terraform so that next time, it can reconcile with infrastructure and get the correct state again. I will keep this as is for now for consistency

sysdig/provider.go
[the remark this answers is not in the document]
ombellare: Done

sysdig/resource_sysdig_secure_notification_channel_msteams.go
ombellare: fixed

sysdig/resource_sysdig_secure_notification_channel_msteams.go
ombellare: Done

ombellare:
Hi @filiptubic, could you please review again? The notification template changes were done by our team in this PR: https://github.com/draios/backend/pull/12557 where we were able to get alerts team approvals. I have tested the terraform steps for all the APIs on my onprem so we should be good :)

sysdig/resource_sysdig_secure_notification_channel_msteams.go
filiptubic: Im not sure this is the logic we want to add because If I try to create a resource, and then read it and if that fails it will remove my resource from state but it will be on API. So I have to import it in order to use it but user will see no warnings nor errors on output. How would user know if he must import something? On the next terraform plan, tool will  probably suggest creating a resource which already exists and that might fail because notification channel with such name already exists...

ombellare: Understood now, I can return the err to the user

filiptubic: Error should be returned but state should not be empty, remove also `d.SetId("")`

ombellare: We are doing it in all current monitor notification channel files as well. Might be good to consult a TF expert on this and remove it everywhere in a separate PR

PR #369, opened 2023-06-14
fix(serverless-agent) Add ignore containers support to TF data source

pgcrooks-sysdig:
# Description
Allow containers to be excluded from instrumentation in the TF data source.

These are currently supported by the CloudFormation `serverless-patcher` so this PR aims to be feature parity to the TF provider.

The tags are parsed from the data source and passed on to the CFN patcher. We map several field names, for example `image`, to their uppercase counterparts. This was extended to include the container `name` field so that the CFN patcher could understand the field.

Dependencies are updated to pull in fixes for Kilt tag parsing from the CFN patcher.


sysdig/data_source_sysdig_fargate_workload_agent.go
```diff
@@ -36,6 +36,13 @@ const agentinoKiltDefinition = `build {
     ]
 }`
 
+var (
+	kiltTagKeys = []string{
```
pgcrooks-sysdig, line 40: Open question - should we allow any arbitrary tags or is it ok to be explicit here?

gnosek: TBH I'd add a separate `ignore_containers` setting that took a list of container names, instead of tightly following the CloudFormation interface. Also, I'm not exactly sure what the use case for ignoring all containers is: just don't pass the task through our instrumentation instead? But if we do want to preserve it, I'd support it as a boolean flag (ignore_all = true/false).

We could then build the tags map ourselves and would have a place to actually add useful documentation (I'm not sure "kilt recipe tags" will be informative to anybody, us after a few months included). It would also sidestep the question you asked here entirely ;)

sysdig/cfn_preprocess_template.go
```diff
@@ -96,6 +96,19 @@ func terraformPreModifications(ctx context.Context, patchedStack []byte) ([]byte
...
+				if err != nil {
+					return nil, fmt.Errorf("Could not update Name field: %v", err)
+				}
+
+				err = container.Delete("name")
+				if err != nil {
+					return nil, fmt.Errorf("could not delete name in the Container definition: %w", err)
+				}
+			}
+
```
gnosek, line 111: Where is this used and what does it do exactly?

sysdig/data_source_sysdig_fargate_workload_agent.go
pgcrooks-sysdig: My main aim was to exactly duplicate the CF interface for this since the documentation is already there an the customer in question tried it. I will ask PM to see what they would prefer, code-wise it's a small delta to change the tags. I think this answers your very valid question about the "ignore all" option, as it is a slightly round-about way to disable instrumentation of the entire task.

pgcrooks-sysdig: @gnosek I spoke to PM - they are happy to copy CF functionality to unblock the customer. I will raise an improvement ticket to review the TF API.

gnosek: I'm not sure it's a good idea to introduce a new feature just to change it again soon. IMHO let's either do it the TF way from the beginning or just leave the internals exposed. Otherwise we'll have to maintain two sets of docs and won't have a simple answer for "so how do I ignore a container in tf?" "well, it depends on your tf provider version"

francesco-racciatti: I'd avoid defining a data struct at the Terraform-level, and rather keep the info related to the container to be ignored at the TaskDefinition-level, as we already do in CFN.
That's for two main reasons:
 - That info relates to the design of the TaskDefinition more than to its deployment.
 - Different containers on different TaskDefinitions could have the same name, so the parameter at the Terraform-level should be a map to identify uniquely the containers to be skipped. In case of a massive deployment that wouldn't be practical.

gnosek: I'm not advocating any changes to how the process works, just an interface change from:
```
kilt_tags = {
  kilt_ignore_containers = "foo:bar"
}
```

to
```
ignore_containers = ["foo", "bar"]
```

(modulo exact hcl syntax).

This shouls all live in the kilt data source. If you need different ignored lists for different workloads, you need specific data sources anyway (there's nowhere to put the tags in the container definition json, since it lives at a higher level).

Note how https://registry.terraform.io/modules/cn-terraform/ecs-fargate-task-definition/aws/latest splits out all the top-level items of the task definition into separate terraform-level variables and we only get the container definition to work with

pgcrooks-sysdig: I can rename my "API" to simplify it, like this simplified example:
```
data "sysdig_fargate_workload_agent" "containers_instrumented" {
  container_definitions = jsonencode([
    {
      "name" : "nginx",
      "image" : "nginx:latest",
...
    ignore_containers = ["nginx"]
  }
}

resource "aws_ecs_task_definition" "task_definition" {
  family             = "${var.prefix}-instrumented-task-definition"
  requires_compatibilities = ["FARGATE"]

  container_definitions = data.sysdig_fargate_workload_agent.containers_instrumented.output_container_definitions
}
```

As @gnosek points out, our data provider does not have the scope to define the "tags" within the task definition itself. I am very happy to drop "support" for the `kilt-ignore` tag as it really does not make much sense here.

francesco-racciatti (approved):
LGTM

francesco-racciatti:
Could you also update the related docs (the `Argument Reference`, in particular) in [website/docs/d/fargate_workload_agent.md](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/SSPROD-24040-fix-access-key-no-longer-required/website/docs/d/fargate_workload_agent.md), pls?

pgcrooks-sysdig:
> Could you also update the related docs (the `Argument Reference`, in particular) in [website/docs/d/fargate_workload_agent.md](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/SSPROD-24040-fix-access-key-no-longer-required/website/docs/d/fargate_workload_agent.md), pls?

Done!

website/docs/d/fargate_workload_agent.md
```diff
@@ -42,6 +42,7 @@ data "sysdig_fargate_workload_agent" "instrumented_containers" {
   * `stream_prefix` - Prefix for the instrumentation log stream
   * `region` - The AWS region where the target log group resides
 * `sysdig_logging` - (Optional) The instrumentation logging level: `trace`, `debug`, `info`, `warning`, `error`, `silent`.
+* `ignore_containers` - (Optional) A list of containers in this data source that should not be instrumented.
```
gnosek, line 45:
```suggestion
* `ignore_containers` - (Optional) A list of containers in the task definition that should not be instrumented.
```

gnosek (commented):
LGTM, modulo the one nit (take it or leave it though).

But also, do we test that both `name` and `Name` work (in user input)?

pgcrooks-sysdig:
> But also, do we test that both `name` and `Name` work (in user input)?

Not explicitly, the closest check is in https://github.com/sysdiglabs/terraform-provider-sysdig/pull/369/files#diff-690c3d8f4a1f27c6bd5d112db02deabb481613e603e2cc3c2b155922df26859aR126
I copied how we tested the `Image` field etc. I can add tests for each of those fields if needed.

francesco-racciatti:
We should already test that, pls see:
 - source [sysdig/testfiles/fargate_field_case_test.json](https://github.com/sysdiglabs/terraform-provider-sysdig/pull/369/files#diff-81125d5e2fe1fd8dde64605927486fd85b4902c47c3a25ce5b7cf0a858c6b121)
 - expected [sysdig/testfiles/fargate_field_case_test_expected.json](https://github.com/sysdiglabs/terraform-provider-sysdig/pull/369/files#diff-c6d884a0e30cfc4b055046ebe07fac3a1ceeba607353037caa837bf4469cd305)

francesco-racciatti (approved):

francesco-racciatti:
Uhh yes eheh, sorry, I hadn't noticed that
[sq case test](https://github.com/sysdiglabs/terraform-provider-sysdig/pull/369/commits/1e4675b293a16c2882aa28ac874a142f2dde9b87)

sysdig/cfn_preprocess_template.go
pgcrooks-sysdig: Resolving via 👍

PR #375, opened 2023-06-30
fix: correct error message for missing secure token

jacalvo:


jacalvo:
closing in favor of #377

PR #389, opened 2023-08-03
feat(notification channel): add new notification channel types

dbonf:
add new notification channel resources for monitor and secure:

* `sysdig_monitor_notification_channel_google_chat`
* `sysdig_monitor_notification_channel_prometheus_alert_manager`
* `sysdig_monitor_notification_channel_team_email`
* `sysdig_monitor_notification_channel_custom_webhook`
* `sysdig_monitor_notification_channel_ibm_event_notification`
* `sysdig_monitor_notification_channel_ibm_function`
* `sysdig_secure_notification_channel_prometheus_alert_manager`
* `sysdig_secure_notification_channel_team_email`


.github/workflows/test.yml
```diff
@@ -105,6 +105,7 @@ jobs:
           SYSDIG_IBM_MONITOR_INSTANCE_ID: ${{ secrets.TERRAFORM_IBM_MONITOR_INSTANCE_ID }}
           SYSDIG_IBM_MONITOR_IAM_URL: "https://iam.cloud.ibm.com"
           SYSDIG_MONITOR_URL: "https://eu-gb.monitoring.cloud.ibm.com"
+          IBM_EVENT_NOTIFICATION_INSTANCE_ID: "973d1834-9731-4c39-8f04-6e860c7e4174"
```
filiptubic, line 108: Should we add this to github envs instead of exposing ID like this? Is it secret?

filiptubic: Even if its not secret I think we should not expose internal IDs liek this

sysdig/resource_sysdig_monitor_notification_channel_custom_webhook.go
```diff
@@ -0,0 +1,187 @@
...
+
+		Schema: createMonitorNotificationChannelSchema(map[string]*schema.Schema{
+			"url": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"http_method": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"POST", "PUT", "PATCH", "DELETE"}, false),
```
filiptubic, line 43: Why we have this validation? Is this validation done on API as well? If yes I would not duplicate code. We do not want to maintain feature parity whenever new thing is added.

sysdig/resource_sysdig_monitor_notification_channel_ibm_cloud_function_test.go
```diff
@@ -0,0 +1,122 @@
...
+	share_with_current_team = true
+}`, name)
+}
+
+func monitorNotificationChannelIBMCloudFunctionWebActionWithCustomData(name string) string {
+	return fmt.Sprintf(`
+resource "sysdig_monitor_notification_channel_ibm_function" "sample4" {
+	name = "Example Channel %s - IBM Function"
+	ibm_function_type = "WEB_ACTION"
+	url = "https://eu-gb.functions.cloud.ibm.com/api/v1/web/namespaces/eeeeeeee-623b-4776-ba35-4065bcbfee7b/actions/hello-world/helloworld?param=true"
```
filiptubic, line 106: Do we need to hardcode url like this? If we switch envs, tests will start failing

filiptubic: Also this will not work on staging?

sysdig/resource_sysdig_monitor_notification_channel_ibm_event_notification_test.go
```diff
@@ -0,0 +1,79 @@
...
+	rText := func() string { return acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum) }
+
+	ibmEventNotificationInstanceId := os.Getenv("IBM_EVENT_NOTIFICATION_INSTANCE_ID")
+	if ibmEventNotificationInstanceId == "" {
+		t.Skip("Skipping tests on sysdig_monitor_notification_channel_ibm_event_notification resource because IBM_EVENT_NOTIFICATION_INSTANCE_ID is not set")
+		return
+	}
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
```
filiptubic, line 27: use `preCheckAnyEnv` func here

filiptubic: also in all other places

sysdig/resource_sysdig_monitor_notification_channel_ibm_cloud_function_test.go
dbonf: urls for ibm functions are not validated, so you can put anything with a precise pattern there and it will work

sysdig/resource_sysdig_monitor_notification_channel_ibm_event_notification_test.go
dbonf: done

filiptubic (approved):
IBM_EVENT_NOTIFICATION_INSTANCE_ID is created under secrets. LGTM just change the github actions to use this new secret.

.github/workflows/test.yml
dbonf: added a secret

PR #394, opened 2023-08-09
fix(posture_zone): exclude username from zone request

chen-shmilovich-sysdig:
Exclude username from `PostureZoneRequest` since it's now automatically added in the backend.


filiptubic (approved):
LGTM but lets wait for tests to pass

PR #404, opened 2023-09-05
SSPROD-29308: Implement CloudAccount client v2 for Secure

ravinadhruve10:
Change summary:
-----------------
Adding a new Cloudauth Account client for making Secure BE (cloudauth) APIs to support API-only onboarding support.


nkraemer-sysdig (commented):
Maybe we can use a `cloudauth` prefix instead of a `v2` suffix across the board? E.g. name the file `sysdig/internal/client/v2/cloudauth.go` and the constants `cloudauthAccountPath`/`cloudauthAccountsPath`?

Seems a bit odd to add `v2`s across the board, especially when refering to an endpoint with `v1` in it

sysdig/internal/client/v2/model.go
```diff
@@ -411,6 +413,35 @@ type CloudAccountSecure struct {
...
+
+type CloudAccountFeatures struct {
+	SecureThreatDetection     *CloudAccountFeature
+	SecureConfigPosture       *CloudAccountFeature
+	SecureIdentityEntitlement *CloudAccountFeature
+	MonitorCloudMetrics       *CloudAccountFeature
+	SecureAgentlessScanning   *CloudAccountFeature
+}
+
+type CloudAccountSecureV2 struct {
```
ravinadhruve10, line 430: Analyzing if we can import and use the Cloud Account object directly from `draios/protorepo/cloudauth`?
And then the `*FromResourceData` and `*ToResourceData` in resource*.go can do the from/to conversion of the fields as required.

ravinadhruve10:
> Maybe we can use a `cloudauth` prefix instead of a `v2` suffix across the board? E.g. name the file `sysdig/internal/client/v2/cloudauth.go` and the constants `cloudauthAccountPath`/`cloudauthAccountsPath`?
> 
> Seems a bit odd to add `v2`s across the board, especially when refering to an endpoint with `v1` in it
> 
> > Yes, that works too and that's a fair point. I named it with v2 as per the framework suggestion, similar to how they did for alerts_v2.go. But cloudauth also works. (as long as we don't change the name of the service).

Done!

sysdig/internal/client/v2/model.go
ravinadhruve10: Done.

ravinadhruve10:
All the latest changes are pulled and merged as part of https://github.com/sysdiglabs/terraform-provider-sysdig/pull/407 . Closing this PR then.

PR #406, opened 2023-09-05
feat(secure-organization): SSPROD-29302 Provider support for organization resource

mdkulkarni15:

### DIFF
--- a/sysdig/internal/client/v2/model.go
+++ b/sysdig/internal/client/v2/model.go
@@ -779,6 +779,7 @@ type SilenceRule struct {
 }
 
 type CloudAccountSecure2 struct {
+	// TODO: Temporarily adding this to support till we add CloudAccountSecure here
 }
 
 type Timestamp struct {
--- a/sysdig/internal/client/v2/model.go
+++ b/sysdig/internal/client/v2/model.go
@@ -777,3 +777,29 @@ type SilenceRule struct {
 	Version int `json:"version,omitempty"`
 	ID      int `json:"id,omitempty"`
 }
+
+type CloudAccountSecure2 struct {
+}
+
+type Timestamp struct {
+	Seconds int64 `json:"seconds,omitempty"`
+	Nanos   int32 `json:"nanos,omitempty"`
+}
+
+type OrganizationSecure struct {
+	Id                    string                 `json:"id,omitempty"`
+	Name                  string                 `json:"name,omitempty"`
+	ManagementAccountId   string                 `json:"management_account_id,omitempty"`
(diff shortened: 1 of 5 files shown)


sysdig/resource_sysdig_secure_organization.go
```diff
@@ -0,0 +1,151 @@
...
+			Delete: schema.DefaultTimeout(timeout),
+		},
+		Schema: map[string]*schema.Schema{
+			"cloud_provider_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"cloud_provider_type": {
+				Type:     schema.TypeInt,
+				Required: true,
```
nkraemer-sysdig, line 35: Should we add validation here to restrict this to just supported providers?

nkraemer-sysdig (commented):
Can we please add some tests?

mdkulkarni15:
> Can we please add some tests?

Yes coming soon.

sysdig/internal/client/v2/model.go
```diff
@@ -777,3 +777,30 @@ type SilenceRule struct {
...
+type CloudAccountSecure2 struct {
+	// TODO: Temporarily adding this to support till we add CloudAccountSecure here
+}
+
+type Timestamp struct {
+	Seconds int64 `json:"seconds,omitempty"`
+	Nanos   int32 `json:"nanos,omitempty"`
+}
+
+type OrganizationSecure struct {
```
cgeers, line 790: Can we just use the [actual type](https://github.com/draios/protorepo/blob/master/cloudauth/go/cloud_organization.pb.go#L24-L39) here? 

if we need to wrap it, or embed it, that still seems better than redefining it.

ravinadhruve10: Yes, I've been thinking about it too, if we can import and use the Cloud Account and Cloud Org object directly from `draios/protorepo/cloudauth`?
And then the `*FromResourceData` and `*ToResourceData` in resource*.go can do the from/to conversion of the fields as required.

mdkulkarni15: The only thing i am not sure about is that, this is a public repo, can we refer internal proto definitions in this? Any implications of exposing the proto to public?  
Well, the api is public facing too, so maybe that's not an issue.

sysdig/resource_sysdig_secure_organization.go
mdkulkarni15: done!

mdkulkarni15:
Yes, will close this.

PR #407, opened 2023-09-06
[SSPROD-29309] - Create a resource for GCP API-only onboarding

raghavanandan:
Change summary:
-----------------
Adding a new CloudAccount resource to support API-only onboarding.


sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -0,0 +1,174 @@
...
+				Required: true,
+			},
+			"cloud_provider_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"cloud_provider_type": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"aws", "gcp", "azure"}, false),
```
mdkulkarni15, line 44: I thought provider type is uint32 defined in the proto, check how it is done here https://github.com/sysdiglabs/terraform-provider-sysdig/pull/408/files#diff-58952c0c98afa44fb3f270b3577f925b263b05328de2b71cbddbb6bd1dea8229

raghavanandan: Updated!

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if err != nil {
+		data.SetId("")
+		return diag.FromErr(err)
+	}
+
+	cloudauthAccount, err := client.GetCloudauthAccountSecure(ctx, data.Id())
+
+	if err != nil {
+		data.SetId("")
```
mdkulkarni15, line 87: Why are we changing data here? Is it necessary?

raghavanandan: I don't think it's necessary. I was following the standard we have for the existing cloud account resource.

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func cloudauthAccountFromResourceData(data *schema.ResourceData) *v2.CloudauthAccountSecure {
+	return &v2.CloudauthAccountSecure{
+		Id:         data.Get("account_id").(string),
```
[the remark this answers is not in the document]
raghavanandan: It's fetched from the resourceData but we're using this function to pass the cloud account object for account creation.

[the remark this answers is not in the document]
ravinadhruve10: Yes, I think the `id` field will be computed from the BE http API and returned back.

- We won't need it here in the `From` func, i.e we won't need this input from the terraform resourceData.
- Adding it in `To` func in L146 looks correct. Since there we are populating in the return path, what we get back as response from the BE http API.

Also, shall we call the resource field name as `account_id` or just `id` in the schema to avoid any confusion? thoughts?

ravinadhruve10 (commented):
Just one comment regarding `account_id`, rest looks good! Thanks @raghavanandan for taking this up!

sysdig/resource_sysdig_secure_cloud_auth_account.go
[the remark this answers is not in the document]
mdkulkarni15: let's keep it as what is defined in proto i.e. Id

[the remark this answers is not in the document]
raghavanandan: The proto differentiates the IDs as `id`, `provider_id` and so on. In that order, I'd prefer having it as `account_id` as it's fairly descriptive and easily distinguishable from `organization_id` or any other ID fields.

sysdig/internal/client/v2/model.go
```diff
@@ -411,6 +415,9 @@ type CloudAccountSecure struct {
 	WorkLoadIdentityAccountID    string `json:"workloadIdentityAccountId,omitempty"`
 	WorkLoadIdentityAccountAlias string `json:"workLoadIdentityAccountAlias,omitempty"`
 }
+
+type CloudauthAccountSecure cloudauth.CloudAccount
```
mdkulkarni15, line 419: Let's not have Cloudauth in the name here, can you rename it to something generic?

sysdig/internal/client/v2/sysdig.go
```diff
@@ -43,6 +43,7 @@ type SysdigSecure interface {
 	VulnerabilityExceptionListInterface
 	VulnerabilityExceptionInterface
 	CloudAccountSecureInterface
+	CloudauthAccountSecureInterface
```
mdkulkarni15, line 46: Same here, let's not have Cloudauth here.

sysdig/internal/client/v2/model.go
ravinadhruve10: Actually, this PR has both my and Raghav's changes. I had renamed files and references to cloudauth in `client` changes from this suggested comment - https://github.com/sysdiglabs/terraform-provider-sysdig/pull/404#pullrequestreview-1613952955

sysdig/internal/client/v2/sysdig.go
ravinadhruve10: same here.

PR #408, opened 2023-09-06
feat(secure): SSPROD-29302 Provider support for organization resource

mdkulkarni15:


sysdig/resource_sysdig_secure_organization.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	return nil
+}
+
+/* TODO: Make sure these are right inputs for the api */
+func secureOrganizationFromResourceData(data *schema.ResourceData) v2.OrganizationSecure {
+	return v2.OrganizationSecure{
+		Id:         data.Get("organization_id").(string),
+		ProviderId: data.Get("cloud_provider_id").(string),
```
ravinadhruve10, line 126: @mdkulkarni15 Do we know what exactly is the `providerId` here? For single account case, I've seen `providerId` set as the `accountId` when we run cloudauth API `/api/cloudauth/v1/accounts` today.
cc: @raghavanandan

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -0,0 +1,154 @@
...
+			Read:   schema.DefaultTimeout(timeout),
+			Delete: schema.DefaultTimeout(timeout),
+		},
+		Schema: map[string]*schema.Schema{
+			"cloud_provider_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"cloud_provider_type": {
+				Type:         schema.TypeInt,
```
raghavanandan, line 36: Even though the provider type is fetched from an enum, the schema expects a string type. Even the single account API's schema expects string type.
https://github.com/draios/secure-backend/blob/main/docs/api/cloudauth/v1/swagger.yaml#L377

mdkulkarni15: we can do a .String() on the enum as below

sysdig/resource_sysdig_secure_organization.go
mdkulkarni15: Right now the api is very aws specific, we need to come up with what fields are required for all the cloud providers. We can discuss this holistically. Id is subscription id/ tenantid/ folderid etc.. providerId is account id in accounts api

ravinadhruve10: Thanks Mandar.  2 questions -
- In that case, this API will work for GCP too then right? Because GCP is the first support we are adding this for.
- Isn't `id` a computed field for `/organizations` API as well (similar to `/accounts` api)

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -0,0 +1,154 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(timeout),
+			Update: schema.DefaultTimeout(timeout),
+			Read:   schema.DefaultTimeout(timeout),
+			Delete: schema.DefaultTimeout(timeout),
+		},
+		Schema: map[string]*schema.Schema{
```
ravinadhruve10, line 30: Based on our offline discussion today, shall we change this in the schema to have the below aligning with the POST `/organizations` API -
- cloud_provider_type (required)
- managementAccountId (required)
- organizationalUnitIds (required)
- id (computed)

We won't need customer_id, organization_id, cloud_provider_id here no?

ravinadhruve10 (commented):
Few comments around the schema fields, rest looks good to me!

sysdig/resource_sysdig_secure_organization.go
mdkulkarni15: let us wait till Haresh makes the organizations api more generic, as i said it is very aws specific. Once we finalize the api this change should be very trivial

sysdig/resource_sysdig_secure_organization.go
mdkulkarni15: Let's revisit this once haresh completes the /organizations refactor. Will get more idea, like managementAccountId and  organizationalUnitIds are AWS remnants might not need for gcp

mdkulkarni15: fixed in the latest push.

sysdig/resource_sysdig_secure_organization.go
mdkulkarni15: Fixed in the latest push.

mdkulkarni15:
Closing this PR as i have accidentally pushed the changes directly to the api-only-secure-onboarding branch.

PR #409, opened 2023-09-08
feat(secure-rule, secure-policy) use status code to determine if resource needs to be recreated

kmvachhani:
We should differentiate between 404 errors and all other errors when trying to get a rule from the backend. If the response is non 404, we should not try to re-create the resource, as it might already exist.


ombellare (commented):
Would the same be required for when a GetPolicyByID returns a 404?

ombellare (approved):
LGTM

PR #412, opened 2023-09-17
add option to skip policyv2msg in policy/rule/list/macro mutating operations

kmvachhani:
when enabled, this new flag will not send a message to the agents with the new contents. applies to all policies/rules/lists/macro creates/deletes/updates

### DIFF
--- a/sysdig/provider.go
+++ b/sysdig/provider.go
@@ -11,6 +11,13 @@ type SysdigProvider struct {
 	SysdigClient SysdigClients
 }
 
+// Used by tests to get the provider
+func Provider() *schema.Provider {
+	sysdigClient := NewSysdigClients()
+	provider := &SysdigProvider{SysdigClient: sysdigClient}
+	return provider.Provider()
+}
+
 func (p *SysdigProvider) Provider() *schema.Provider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
--- a/sysdig/provider.go
+++ b/sysdig/provider.go
@@ -10,6 +10,11 @@ import (
 func Provider() *schema.Provider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
+			"sysdig_secure_skip_policyv2msg": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SYSDIG_SECURE_SKIP_POLICYV2MSG", false),
+			},
 			"sysdig_secure_api_token": {
 				Type:        schema.TypeString,
 				Optional:    true,
--- a/sysdig/provider.go
+++ b/sysdig/provider.go
@@ -7,13 +7,17 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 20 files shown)


ben-m-lucas (approved):
Looks good.  Thank you for working on this.

ivanlysiuk-sysdig (approved):

sysdig/provider.go
```diff
@@ -7,9 +7,25 @@ import (
...
+	return provider.Provider()
+}
+
+func (p *SysdigProvider) Provider() *schema.Provider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
+			"sysdig_secure_skip_policyv2msg": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SYSDIG_SECURE_SKIP_POLICYV2MSG", true),
```
ombellare, line 27: So by default we will not send PV2Ms on terraform changes?

sysdig/resource_sysdig_secure_policy.go
```diff
@@ -335,5 +343,31 @@ func resourceSysdigPolicyUpdate(ctx context.Context, d *schema.ResourceData, met
...
 		return diag.FromErr(err)
 	}
+	sysdigClients.AddCleanupHook(sendPoliciesToAgents)
+
 	return nil
 }
+
+var sendPoliciesToAgentsOnce sync.Once
+
+func sendPoliciesToAgents(ctx context.Context, clients SysdigClients) error {
```
ombellare, line 353: So if there are 10 rule/macro/list changes, how would the behavior be after this change?

sysdig/sysdig_clients.go
```diff
@@ -159,6 +186,26 @@ func getIBMSecureVariables(data *schema.ResourceData) (*ibmVariables, error) {
...
 }
 
+func (c *sysdigClients) Configure(ctx context.Context, d *schema.ResourceData) {
+	c.ctx = ctx
+	c.d = d
+}
+
+func (c *sysdigClients) AddCleanupHook(cleanupHook func(context.Context, SysdigClients) error) {
+	c.mu.Lock()
+	c.cleanupHooks = append(c.cleanupHooks, cleanupHook)
```
ombellare, line 196: What would happen if we add multiple cleanup hooks of the same type, in this case the sendPoliciesToAgents?

ombellare (commented):
Just a few clarifying comments

filiptubic (approved):
I've already reviewed Ben's PR. Regarding those changes LGTM.

sysdig/provider.go
ben-m-lucas: That's correct.  They get sent at the end.

sysdig/resource_sysdig_secure_policy.go
ben-m-lucas: It will make one call to the API to send the policies to agents.
That's the purpose of using sync.Once.  It guarantees that we call this only once.

ombellare (approved):
Thanks for answering my questions. LGTM! 💯

PR #415, opened 2023-09-20
Forward PolicyV2 messages after all changes are complete

ben-m-lucas:
This PR:
sets up a hook that would allow us to call the backend API to forward policyv2 messages after all of the changes are complete.

Here's a sample log of what this looks like running locally:

```
blucas@ben-lucas:~/tests/terraform-demo$ terraform plan
2023-09-20T14:38:47.074-0700 [WARN]  ValidateProviderConfig from "provider[\"local/sysdiglabs/sysdig\"]" changed the config value, but that value is unused
sysdig_secure_rule_falco.common_netcat_remote_code_execution_in_container: Refreshing state... [id=10002534]
2023-09-20T14:38:47.531-0700 [WARN]  Provider "local/sysdiglabs/sysdig" produced an invalid plan for sysdig_secure_rule_falco.common_netcat_remote_code_execution_in_container, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .condition: planned value cty.StringVal("") for a non-computed attribute
      - .source: planned value cty.StringVal("") for a non-computed attribute
      - .description: planned value cty.StringVal("") for a non-computed attribute
      - .output: planned value cty.StringVal("") for a non-computed attribute
      - .priority: planned value cty.StringVal("warning") for a non-computed attribute
2023-09-20T14:38:47.532-0700 [WARN]  provider.terraform-provider-sysdig: [WARN] Closing sysdig clients - check for need to send request to forward policyv2: @module=provider tf_provider_addr=provider tf_req_id=3433bfda-1d40-ff9b-f5f3-b602ddc141f3 tf_rpc=Configure @caller=/home/blucas/repos/terraform-provider-sysdig/sysdig/sysdig_clients.go:193 timestamp=2023-09-20T14:38:47.532-0700

```

The intent would be to modify sysdigClients to track mutating calls to rules, lists, macros, and policies and if any have been made, it would make an API call to forward the PolicyV2 message at the end.

### DIFF
--- a/sysdig/internal/client/v2/client.go
+++ b/sysdig/internal/client/v2/client.go
@@ -65,9 +65,8 @@ type Requester interface {
 }
 
 type Client struct {
-	config          *config
-	requester       Requester
-	policiesChanged bool
+	config    *config
+	requester Requester
 }
 
 func (client *Client) ErrorFromResponse(response *http.Response) error {
--- a/sysdig/internal/client/v2/client.go
+++ b/sysdig/internal/client/v2/client.go
@@ -65,8 +65,9 @@ type Requester interface {
 }
 
 type Client struct {
-	config    *config
-	requester Requester
+	config          *config
+	requester       Requester
+	policiesChanged bool
 }
 
 func (client *Client) ErrorFromResponse(response *http.Response) error {
--- a/sysdig/internal/client/v2/list.go
+++ b/sysdig/internal/client/v2/list.go
@@ -37,8 +37,6 @@ func (client *Client) CreateList(ctx context.Context, list List) (List, error) {
 		return List{}, client.ErrorFromResponse(response)
 	}
 
-	client.policiesChanged = true
-
 	return Unmarshal[List](response.Body)
 }
 
@@ -81,8 +79,6 @@ func (client *Client) UpdateList(ctx context.Context, list List) (List, error) {
 		return List{}, client.ErrorFromResponse(response)
 	}
 
-	client.policiesChanged = true
-
 	return Unmarshal[List](response.Body)
 }
 
@@ -96,8 +92,6 @@ func (client *Client) DeleteList(ctx context.Context, id int) error {
 	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
 		return client.ErrorFromResponse(response)
 	}
-
-	client.policiesChanged = true
 
 	return nil
 }
--- a/sysdig/internal/client/v2/list.go
+++ b/sysdig/internal/client/v2/list.go
@@ -37,6 +37,8 @@ func (client *Client) CreateList(ctx context.Context, list List) (List, error) {
 		return List{}, client.ErrorFromResponse(response)
 	}
 
+	client.policiesChanged = true
+
 	return Unmarshal[List](response.Body)
 }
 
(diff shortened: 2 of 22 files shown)


sysdig/sysdig_clients.go
```diff
@@ -173,6 +184,20 @@ func getIBMSecureVariables(data *schema.ResourceData) (*ibmVariables, error) {
...
+}
+
+func (c *sysdigClients) Close() error {
+	ctx, _ := context.WithDeadline(context.Background(), time.Now().Add(15*time.Second))
+	err := c.secureClientV2.SendPoliciesToAgents(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
filiptubic, line 199: Here we are adding code specific to policies. Sysdig clients is used for other parts of the sysdig, for monitor as well. What will happen if user authenticates only on monitor. secureClientV2 will be nil. Even if its not nil I would not like to have in this code anything releated to specific feature on our platform

filiptubic: This file and SysdigClient is used to create clients depending on the product and environment not to perform actions on policies or any other feature.

sysdig/internal/client/v2/client.go
```diff
@@ -65,8 +65,9 @@ type Requester interface {
 }
 
 type Client struct {
-	config    *config
-	requester Requester
+	config          *config
+	requester       Requester
+	policiesChanged bool
```
filiptubic, line 70: Again we have specific code related to policies in Client which is completely abstract.

sysdig/internal/client/v2/list.go
```diff
@@ -93,6 +97,8 @@ func (client *Client) DeleteList(ctx context.Context, id int) error {
 		return client.ErrorFromResponse(response)
 	}
 
+	client.policiesChanged = true
```
filiptubic, line 100: Why not calling here endpoint that policy is changed? Also same question for other places?

filiptubic (changes requested):
I've added few comments. Instead of adding this Close function, each resource handler must be self sufficient. After modifying the policy resource we should call there in place endpoint for sending the message. But we should not to something like this, marking it as dirty and then send message when client is closed.

ben-m-lucas:
> I've added few comments. Instead of adding this Close function, each resource handler must be self sufficient. After modifying the policy resource we should call there in place endpoint for sending the message. But we should not to something like this, marking it as dirty and then send message when client is closed.

In this case, the problem we are trying to solve is a flood of messages to the agent.  The idea is that we only want to call this once.  If we call it for each mutation, that defeats the purpose of this PR.

jacklongsd (approved):

sysdig/internal/client/v2/list.go
ben-m-lucas: Modified to move logic for whether to call the endpoint into the related resources

sysdig/internal/client/v2/client.go
ben-m-lucas: Modified to move policies specific code out of the abstract objects.

sysdig/sysdig_clients.go
ben-m-lucas: Modified per our discussion to create a generic concept of cleanup hooks that can be used by other features as well.

miguelgordo (approved):

filiptubic (approved):
Thanks for updating with cleanup functions. LGTM

PR #417, opened 2023-09-21
fix(secure-cloud-account): Support parsing JSON encoded string inputs for Component metadata

ravinadhruve10:
Current (and working) TF snippet expected from `onboarding-api` :-

```
terraform {
  required_providers {
    sysdig = {
      source = "local/sysdiglabs/sysdig"
      version = "~> 1.0.0"
    }
  }
}

provider "sysdig" {
  sysdig_secure_url          = "https://secure-staging.sysdig.com"
  sysdig_secure_api_token    = "fgb6xxxx-xxxxx-xxxxx-xxxxxxxx"
}

provider "google" {
  project = "mygcpproject"
  region  = "us-west1"
}

module "single-project-cspm" {
  source               = "sysdiglabs/secure/google//modules/services/service-principal"
  project_id           = "mygcpproject"
  service_account_name = "ravina-test-sa"
}

resource "sysdig_secure_cloud_auth_account" "gcp_account_provider_id" {
  provider_id   = "mygcpproject"
  provider_type = "PROVIDER_GCP"
  enabled       = true
  feature {
      secure_config_posture {
        enabled    = "true"
        components = ["COMPONENT_SERVICE_PRINCIPAL/secure-service-principal"]
      }
  }
  component {
      type                       = "COMPONENT_SERVICE_PRINCIPAL"
      instance                   = "secure-service-principal"
      service_principal_metadata = "{\"gcp\":{\"key\":\"${module.single-project-cspm.service_account_key}\"}}"
  }
}
```
**Some points to note:**

- The provider source will change ofcourse once released.
- Both below formats for `service_principal_metadata` work currently. (targeting the latter during TF snippet generation which is more readable)
```
service_principal_metadata = "{\"gcp\":{\"key\":\"${module.single-project-cspm.service_account_key}\"}}"
```
 and
```
      service_principal_metadata = jsonencode({
        gcp = {
          key = module.single-project-cspm.service_account_key
        }
      })
```

### DIFF
--- a/sysdig/resource_sysdig_secure_cloud_auth_account.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account.go
@@ -343,19 +343,26 @@ func constructAccountComponents(accountComponents []*cloudauth.AccountComponent,
 					}
 				case SchemaServicePrincipalMetadata:
 					// TODO: Make it more generic than just for GCP
-					service_principal_private_key := getServicePrincipalKeyObject(value.(string))
-					component.Metadata = &cloudauth.AccountComponent_ServicePrincipalMetadata{
-						ServicePrincipalMetadata: &cloudauth.ServicePrincipalMetadata{
-							Provider: &cloudauth.ServicePrincipalMetadata_Gcp{
-								Gcp: &cloudauth.ServicePrincipalMetadata_GCP{
-									Key: &cloudauth.ServicePrincipalMetadata_GCP_Key{
-										ProjectId:    data.Get(SchemaCloudProviderId).(string),
-										PrivateKeyId: service_principal_private_key["private_key_id"],
-										PrivateKey:   service_principal_private_key["private_key"],
+					servicePrincipalMetadata := parseMetadataJson(value.(string))
+					if provider == cloudauth.Provider_PROVIDER_GCP.String() {
+						encodedServicePrincipalKey, ok := servicePrincipalMetadata["gcp"].(map[string]interface{})["key"].(string)
+						if !ok {
+							fmt.Printf("Component metadata for provider %s is invalid and not as expected", provider)
+						}
+						servicePrincipalKey := getGcpServicePrincipalKey(encodedServicePrincipalKey)
+						component.Metadata = &cloudauth.AccountComponent_ServicePrincipalMetadata{
+							ServicePrincipalMetadata: &cloudauth.ServicePrincipalMetadata{
+								Provider: &cloudauth.ServicePrincipalMetadata_Gcp{
+									Gcp: &cloudauth.ServicePrincipalMetadata_GCP{
+										Key: &cloudauth.ServicePrincipalMetadata_GCP_Key{
+											ProjectId:    data.Get(SchemaCloudProviderId).(string),
+											PrivateKeyId: servicePrincipalKey["private_key_id"],
+											PrivateKey:   servicePrincipalKey["private_key"],
+										},
 									},
 								},
 							},
-						},
+						}
 					}
 				case SchemaWebhookDatasourceMetadata:
 					component.Metadata = &cloudauth.AccountComponent_WebhookDatasourceMetadata{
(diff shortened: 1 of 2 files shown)


sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -130,11 +130,11 @@ func resourceSysdigSecureCloudauthAccount() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
-			"cloud_provider_id": {
+			"provider_id": {
```
mdkulkarni15, line 133: do you want to have them defined as constants above?

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -343,19 +343,26 @@ func constructAccountComponents(accountComponents []*cloudauth.AccountComponent,
...
-					service_principal_private_key := getServicePrincipalKeyObject(value.(string))
-					component.Metadata = &cloudauth.AccountComponent_ServicePrincipalMetadata{
-						ServicePrincipalMetadata: &cloudauth.ServicePrincipalMetadata{
-							Provider: &cloudauth.ServicePrincipalMetadata_Gcp{
-								Gcp: &cloudauth.ServicePrincipalMetadata_GCP{
-									Key: &cloudauth.ServicePrincipalMetadata_GCP_Key{
-										ProjectId:    data.Get("cloud_provider_id").(string),
-										PrivateKeyId: service_principal_private_key["private_key_id"],
-										PrivateKey:   service_principal_private_key["private_key"],
+					servicePrincipalMetadata := parseMetadataJson(value.(string))
```
mdkulkarni15, line 346: maybe you can refactor this and take provider as input for a new function?

mdkulkarni15:
terraform {
  required_providers {
    sysdig = {
      source = "local/sysdiglabs/sysdig"
      version = "~> 1.0.0"
    }
  }
}

Does this need to go to versions.tf? I don't think we need to print it during snippet generation.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: So, the way I have written this `parseMetadataJson` function is - its a common generic function which parses any metadata jsonified string and returns a map. For instance, you can even pass `trusted_role_metadata` to it and it will return a map of key:value pairs.

If I pass in the `provider` as well, then I'll also need to know which metadata I'm parsing for. This is because the metadata are different for different providers. We will likely end up having the same checks for parsing as this switch case, unless ofcourse we have different parsing functions for different metadata types.
Please let me know if I misunderstood the comment.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: Keeping with the convention in the provider framework as other resources, I think its fine. Besides do you see any value in defining them as constants?

ravinadhruve10:
> terraform { required_providers { sysdig = { source = "local/sysdiglabs/sysdig" version = "~> 1.0.0" } } }
> 
> Does this need to go to versions.tf? I don't think we need to print it during snippet generation.

I first did try having it only in versions.tf. But terraform wasn't able to initialize and load the sysdig provider.
Looked at the legacy snippets generated, apparently we do need to add sysdig to the required_providers in the main snippet ran by the user.

sysdig/resource_sysdig_secure_cloud_auth_account.go
mdkulkarni15: well sysdig/common.go technically should have these, but i don't see all of the resources followed the convention. Upto you

sysdig/resource_sysdig_secure_cloud_auth_account.go
mdkulkarni15: I feel we will end up in if/else maze in this switch case. For now it's ok but maybe we want to revisit this.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: Ah I see. Will take a look and update the PR.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: Yeah, I agree. I was equally concerned what if we end up having the if/else/switch maze in the new function too. Let me rethink through this and take it up later if we need to refactor.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: Done.

mdkulkarni15 (approved):

PR #422, opened 2023-09-26
Add Cloudauth Account resource Documentation

ravinadhruve10:

### DIFF
--- a/website/docs/r/secure_cloud_auth_account.md
+++ b/website/docs/r/secure_cloud_auth_account.md
@@ -24,7 +24,7 @@ resource "sysdig_secure_cloud_auth_account" "sample" {
 
 ## Argument Reference
 
-* `provider_id` - (Required) The unique identifier of the cloud account. e.g. for GCP: `mygcpproject`,
+* `provider_id` - (Required) The unique identifier of the cloud account. e.g. for GCP: `mygcpproject`.
 
 * `provider_type` - (Required) The cloud provider in which the account exists. Currently supported provider is `PROVIDER_GCP`.
 
@@ -33,6 +33,8 @@ resource "sysdig_secure_cloud_auth_account" "sample" {
 * `feature` - (Optional) The name and configuration of each feature along with the respective components to enable on this cloud account.
 
 * `component` - (Optional) The component configuration to enable on this cloud account. There can be multiple component blocks for a feature, one for each component to be enabled.
+
+-> **Note:** Please refer to Sysdig Secure API Documentation for the Cloud Accounts API for providing `feature` & `component`.
 
 ## Attributes Reference
 
--- a/website/docs/r/secure_cloud_auth_account.md
+++ b/website/docs/r/secure_cloud_auth_account.md
@@ -1,0 +1,54 @@
+---
+subcategory: "Sysdig Secure"
+layout: "sysdig"
+page_title: "Sysdig: sysdig_secure_cloud_auth_account"
+description: |-
+  Creates a Sysdig Secure Cloud Account using Cloudauth APIs.
+---
+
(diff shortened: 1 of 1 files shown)


website/docs/r/secure_cloud_auth_account.md
```diff
@@ -0,0 +1,54 @@
...
+
+* `provider_id` - (Required) The unique identifier of the cloud account. e.g. for GCP: `mygcpproject`,
+
+* `provider_type` - (Required) The cloud provider in which the account exists. Currently supported provider is `PROVIDER_GCP`.
+
+* `enabled` - (Required) Whether or not to enable sysdig provisioning of resources on this cloud account.
+
+* `feature` - (Optional) The name and configuration of each feature along with the respective components to enable on this cloud account.
+
+* `component` - (Optional) The component configuration to enable on this cloud account. There can be multiple component blocks for a feature, one for each component to be enabled.
```
mdkulkarni15, line 50: do we want to provide an hcl for component  and feature?

raghavanandan: Also, is it standard to mention the types of these params?

ravinadhruve10: I've been on the fence for this one. Based on the conventions I saw for the existing provider resources, its been common to mention all parameters,`required` & `optional` in the HCL. As per my understanding it is to provide a full view of what and how all the fields can be provided for a particular resource.

Having said that, for our resource, there are 2 reasons why we may not want to :-
1. Since we are exposing this resource publicly, we may not want to expose details of `feature` & `component`. However, that means we are documenting how to create a basic cloud account using just provider and account info.  For our resource, the `feature` & `component` is actually not any less public information than the cloudauth APIs themselves.

2. Exposing it also means that the user needs to know what feature and components mapping needs to be provided in the HCL. Since this resource is used by our internal services, we know of the mapping. However someone consuming it externally would need to have this knowledge.

thoughts?

@raghavanandan - param types are implicit in the example usage. I don't think we need to mention them explicitly.

website/docs/r/secure_cloud_auth_account.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Attributes Reference
```
ravinadhruve10, line 52: @mdkulkarni15 Are we planning to allow TF imports for `sysdig_secure_cloud_auth_account` and `sysdig_secure_organization` resources?

mdkulkarni15: Not yet, but maybe in the future.

website/docs/r/secure_cloud_auth_account.md
mdkulkarni15: For point 1 atleast we can link to the structure which is documented should be sufficient. The user should ideally know what and how to populate the feature/component section. Essentially we want to expose the entire payload structure for the cloud account api which is already public.

ravinadhruve10: We can mention all the arguments without exposing any details on the feature/component section, for example [google_service_account resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account). I've updated the PR, is it better?

By _'link to the structure which is documented'_, are we talking about link to the cloudaccount API docs?

mdkulkarni15: yes, if we can add a link to cloud account api for payload that should be fine i assume.

ravinadhruve10: Okay yeah should be fine, let me check the link.

mdkulkarni15 (approved):

PR #437, opened 2023-10-26
feat(cloudauth): organizational units

cgeers:
* add `organizational_unit_ids` to the secure organization resource

### DIFF
--- a/sysdig/common.go
+++ b/sysdig/common.go
@@ -60,4 +60,5 @@ const (
 	SchemaFeature                   = "feature"
 	SchemaManagementAccountId       = "management_account_id"
 	SchemaOrganizationIDKey         = "organization_id"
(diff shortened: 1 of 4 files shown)


ravinadhruve10 (approved):
Overall the change lgtm. Good to have another set of eyes review it. 
Were we able to test it that it works and has no regressions?

ravinadhruve10:
> Overall the change lgtm. Good to have another set of eyes review it. Were we able to test it that it works and has no regressions?

Missed out on this. we will need a documentation change [here](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/master/website/docs/r/secure_organization.md) with this new optional schema field.

ravinadhruve10 (changes requested):
Can we add the documentation change as well as part of this PR?

ravinadhruve10 (approved):

PR #440, opened 2023-10-28
Add support for Secure policy capture fields

ombellare:
This PR adds support for the following three Secure Policy capture related optional fields:
- Filter (Filter string to reduce scope of a capture, eg. `proc.name=vi`)
- Bucket Name (to provide a custom S3 bucket to store the capture files)
- Folder (to provide a folder inside the S3 bucket)


kmvachhani (approved):
lgtm!

PR #442, opened 2023-11-01
chore(serverless-agent) Rename Sysdig agent access key

pgcrooks-sysdig:
Rename to sysdig_agent_access_key to match BE UI.


pgcrooks-sysdig:
Terrible idea that would change the world. Closing.

PR #445, opened 2023-11-08
Implement cloud account creation for Azure

raghavanandan:
Add support for service principal metadata for Azure provider.


sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -413,6 +417,30 @@ func constructAccountComponents(accountComponents []*cloudauth.AccountComponent,
...
+							break
+						}
+
+						component.Metadata = &cloudauth.AccountComponent_ServicePrincipalMetadata{
+							ServicePrincipalMetadata: &cloudauth.ServicePrincipalMetadata{
+								Provider: &cloudauth.ServicePrincipalMetadata_Azure_{
+									Azure: &cloudauth.ServicePrincipalMetadata_Azure{
+										ActiveDirectoryServicePrincipal: &cloudauth.ServicePrincipalMetadata_Azure_ActiveDirectoryServicePrincipal{
+											AccountEnabled:         servicePrincipalAzureKey["account_enabled"].(bool),
+											AppDisplayName:         servicePrincipalAzureKey["app_display_name"].(string),
```
haresh-suresh, line 434: What happens if any of these fields are empty / nil ?

haresh-suresh:
LGTM overall, do we need to add some tests to validate the api response parsing ?
In that vein, do we need acceptance tests for azure single subscription ?

haresh-suresh:
LGTM but wanted to double check whether nil / empty on any of these api response fields will be handled gracefully ?

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -413,6 +417,30 @@ func constructAccountComponents(accountComponents []*cloudauth.AccountComponent,
 								},
 							},
 						}
+					} else if provider == cloudauth.Provider_PROVIDER_AZURE.String() {
+						servicePrincipalAzureKey, ok := servicePrincipalMetadata["azure"].(map[string]interface{})["active_directory_service_principal"].(map[string]interface{})
+
+						if !ok {
+							fmt.Printf("Resource input for component metadata for provider %s is invalid and not as expected\n\n\n", provider)
```
ravinadhruve10, line 424: nit: multiple \n by mistake?

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -159,6 +159,10 @@ func resourceSysdigSecureCloudauthAccount() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			SchemaCloudProviderTenantId: {
+				Type:     schema.TypeString,
+				Required: true,
```
ravinadhruve10, line 164: 2 questions here :-
1. Don't we need to do a data.Get() and data.Set() for `SchemaCloudProviderTenantId` field in From* and To* methods?
2. Since this is a `required` field, what will be the value passed for `provider_tenant_id` in case of other providers such as GCP?

mdkulkarni15: This is only required for Azure, i don't think we can mark it as required for all vendors. Can we possibly validate after receiving the inputs?

ravinadhruve10: Yes if we mark it as `Required` but don't supply the field in case of other vendors, TF itself would complain. Like you mentioned, instead we can validate and enforce it after receiving the inputs. We can validate it (for Azure) while doing `data.Get()` within the `*FromResourceData()` method.

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -650,6 +656,50 @@ func componentsToResourceData(components []*cloudauth.AccountComponent, dataComp
...
+					var out bytes.Buffer
+					if err := json.Indent(&out, bytesKey, "", "  "); err != nil {
+						fmt.Printf("Failed to populate %s: %v", SchemaServicePrincipalMetadata, err)
+						break
+					}
+					out.WriteByte('\n')
+
+					schema, err := json.Marshal(map[string]interface{}{
+						"azure": map[string]interface{}{
+							"key": out.Bytes(),
```
ravinadhruve10, line 693: I believe the structure here is different for azure? instead of `key` it will be `active_directory_service_principal`

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: @raghavanandan lets change this to Optional? now that you are taking care of reading this value only for Azure in L525

sysdig/resource_sysdig_secure_cloud_auth_account.go
raghavanandan: Good catch!

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: Thanks!

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -622,6 +656,50 @@ func componentsToResourceData(components []*cloudauth.AccountComponent, dataComp
...
+					}{
+						AccountEnabled:         providerKey.Azure.GetActiveDirectoryServicePrincipal().GetAccountEnabled(),
+						AppDisplayName:         providerKey.Azure.GetActiveDirectoryServicePrincipal().GetAppDisplayName(),
+						AppId:                  providerKey.Azure.GetActiveDirectoryServicePrincipal().GetAppId(),
+						AppOwnerOrganizationId: providerKey.Azure.GetActiveDirectoryServicePrincipal().GetAppOwnerOrganizationId(),
+						DisplayName:            providerKey.Azure.GetActiveDirectoryServicePrincipal().GetDisplayName(),
+						Id:                     providerKey.Azure.GetActiveDirectoryServicePrincipal().GetId(),
+					}
+
+					bytesKey, err := json.Marshal(jsonifiedKey)
```
ravinadhruve10, line 677: nit/non-blocking: common code in L677 - L689 can be restructured to a helper function

ravinadhruve10 (approved):
Looks like CI complains of some fmt and lint issue. But overall the change LGTM.
I agree adding some acceptance tests will help, that can be part of same / separate PR.

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -512,21 +512,16 @@ func cloudauthAccountFromResourceData(data *schema.ResourceData) *v2.CloudauthAc
...
 	return &v2.CloudauthAccountSecure{
-		CloudAccount: *cloudAccount,
+		CloudAccount: cloudauth.CloudAccount{
+			Enabled:          data.Get(SchemaEnabled).(bool),
+			OrganizationId:   data.Get(SchemaOrganizationIDKey).(string),
+			ProviderId:       data.Get(SchemaCloudProviderId).(string),
+			Provider:         cloudauth.Provider(cloudauth.Provider_value[data.Get(SchemaCloudProviderType).(string)]),
+			Components:       accountComponents,
+			Feature:          accountFeatures,
+			ProviderTenantId: data.Get(SchemaCloudProviderTenantId).(string),
```
ravinadhruve10, line 523: Actually yes this makes sense. In case of other vendors, this will just be passed as empty. For Azure, if this is empty, BE will take care of returning a bad request for missing `provider_tenant_id`. And it will keep the provider lean/dumb.

raghavanandan: That's right! However, I'm conditionally adding this field in the `toResourceData`.

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -153,3 +153,102 @@ func getEncodedServiceAccountKey(resourceName string, accountID string) string {
...
+		type                       = "COMPONENT_SERVICE_PRINCIPAL"
+		instance                   = "secure-posture"
+		service_principal_metadata = jsonencode({
+			azure = {
+				active_directory_service_principal = {
+					id = "%s"
+					account_enabled = true
+					display_name = "test-name"
+					app_display_name = "test-app-name"
+					app_id = %s
```
mdkulkarni15, line 247: quotes here and below?

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -153,3 +153,102 @@ func getEncodedServiceAccountKey(resourceName string, accountID string) string {
...
+func secureAzureCloudAuthAccountWithFC(accountID string) string {
+	rID := func() string { return acctest.RandStringFromCharSet(36, acctest.CharSetAlphaNum) }
+	randomId := rID()
+	randomAppId := rID()
+	randomAppOrgId := rID()
+	return fmt.Sprintf(`
+resource "sysdig_secure_cloud_auth_account" "sample-1" {
+	  provider_id   = "sample-1-%s"
+	  provider_type = "PROVIDER_AZURE"
+	  enabled       = "true"
```
mdkulkarni15, line 230: Do you need quotes here? and below L233?

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -153,3 +153,102 @@ func getEncodedServiceAccountKey(resourceName string, accountID string) string {
...
+		},
+	})
+}
+
+func secureCloudAuthAccountMinimumConfigurationAzure(accountId string) string {
+	return fmt.Sprintf(`
+resource "sysdig_secure_cloud_auth_account" "sample" {
+	  provider_id   = "sample-%s"
+	  provider_type = "PROVIDER_AZURE"
+	  enabled       = "true"
```
mdkulkarni15, line 189: Probably don't need quotes here as well.

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -153,3 +153,102 @@ func getEncodedServiceAccountKey(resourceName string, accountID string) string {
...
+}
+
+func secureAzureCloudAuthAccountWithFC(accountID string) string {
+	rID := func() string { return acctest.RandStringFromCharSet(36, acctest.CharSetAlphaNum) }
+	randomId := rID()
+	randomAppId := rID()
+	randomAppOrgId := rID()
+	return fmt.Sprintf(`
+resource "sysdig_secure_cloud_auth_account" "sample-1" {
+	  provider_id   = "sample-1-%s"
```
mdkulkarni15, line 228: you probably need provider_tenant_id here

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -153,3 +153,102 @@ func getEncodedServiceAccountKey(resourceName string, accountID string) string {
...
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func secureCloudAuthAccountMinimumConfigurationAzure(accountId string) string {
+	return fmt.Sprintf(`
+resource "sysdig_secure_cloud_auth_account" "sample" {
+	  provider_id   = "sample-%s"
```
mdkulkarni15, line 187: do you need provider_tenant_id here?

raghavanandan: Added!

PR #446, opened 2023-11-09
(secure-onboarding) Enable acceptance test for Secure cloud account

ravinadhruve10:


haresh-suresh (commented):
The change itself LGTM, but looks like `govet` wants you to correctly return a string in case of errors too 
```
if err != nil {
     error := fmt.Errorf(...)
     return error.Error()
}
```

haresh-suresh (approved):

PR #448, opened 2023-11-16
Pass provider alias to cloud account creation call

raghavanandan:
This PR is to pass `provider_alias` field as part of cloud account creation call to support org onboarding.


mdkulkarni15 (approved):
LGTM

haresh-suresh (approved):

haresh-suresh:
Does any test need to be updated because of the new field for validation ?

PR #457, opened 2023-12-06
feat(cloud-auth): Add event hub metadata in the event bridge metadata for the azure case

matteopasa:
This PR adds event hub metadata in the cloud auth account resource, needed when onboarding azure accounts.
The cloud auth proto has been updated in the protorepo and copy pasted here.


sysdig/internal/client/v2/cloudauth/go/cloud_account.pb.go
ravinadhruve10, line 1: I think it will be good to always try updating both `pb.go` files together - `cloud_account.pb.go` and `cloud_organization.pb.go`, to avoid any disparities. If `cloud_organization.pb.go` doesn't have any diff, then that's totally okay.

ravinadhruve10 (commented):
One comment, rest lgtm!

PR #461, opened 2023-12-13
Add secure-onboarding organization acc test

ravinadhruve10:
Change summary:
--------------------
[As per TF documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests) - Terraform acceptance tests use real Terraform configurations to exercise the code in real plan, apply, refresh, and destroy life cycles.

TF acceptance tests for secure org onboarding **need an actual existing gcp project** along **with an actual service_principal_key** to scrape all folders and projects under the org. Without it POST /organizations API fail with 500 error and hence we could not add org acceptance tests working with the sysdig backend APIs.

To handle this, adding the org acceptance tests with skip on this particular error check. With this, we can get some decent test coverage as a start :-
- that a valid HCL gets called to the TF sysdig provider, and
- that the right organization API gets called to the sysdig BE.


sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+		},
+		ProviderFactories: map[string]func() (*schema.Provider, error){
+			"sysdig": func() (*schema.Provider, error) {
+				return sysdig.Provider(), nil
+			},
+		},
+		ErrorCheck: func(err error) error {
+			// if regex matches with the expected error, do t.Skip
+			re := regexp.MustCompile(fmt.Sprintf(`POST %s giving up after 5 attempt(s)`, organizationApiUrl))
```
mdkulkarni15, line 41: are there any other error paths which need to be handled here?

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -0,0 +1,132 @@
...
+	  enabled    = "true"
+	  components = ["COMPONENT_SERVICE_PRINCIPAL/secure-posture"]
+	}
+	secure_identity_entitlement {
+	  enabled    = true
+	  components = ["COMPONENT_SERVICE_PRINCIPAL/secure-posture"]
+	}
+  }
+  component {
+	type                       = "COMPONENT_SERVICE_PRINCIPAL"
```
mdkulkarni15, line 81: nit: the spacing looks off compared to below L89-L96

mdkulkarni15 (approved):

sysdig/resource_sysdig_secure_organization_test.go
ravinadhruve10: Interesting, it didn't show up in my IDE. Fixed it in next commit. Thanks :)

sysdig/resource_sysdig_secure_organization_test.go
ravinadhruve10: Not that I'm aware of. Besides we would want the test to catch other errors, such as 400 Bad Request etc.
Anything else you think we should handle?

ravinadhruve10: Resolving this. If required to add handling of specific errors, it can be a follow-up PR.

PR #471, opened 2024-01-13
fix(secure-onboarding) Add WebhookDatasource metadata for GCP case

ravinadhruve10:
- Updated the pb.go files
- Added the webhook datasource metadata case for GCP cloud accounts
- Added corresponding TF ACC tests


ravinadhruve10:
The TF acceptance tests pass when ran against staging, hence the CI tests here are not failing due to code changes in the PR. They can pass only after the proto files used by cloudauth BE are released to all environments, including prod. Moving this to Draft till then.

ravinadhruve10:
This change is no longer needed in the light of recent changes in approach. Closing this.

PR #481, opened 2024-02-19
test(cloud-auth-account): ignores data order and removes the ImportStateVerifyIgnore

wideawakening:
test to understand ordering requirement and why are we ignoring it on the `ImportStateVerify`.
when not adding the `ImportStateVerifyIgnore` tests fail not because of order, but because we cannot retrieve from the state while setting the state. added a fallback for when no order is available.

if order is important we should add a test to validate that use-case, but not able to reproduce it right now.


ravinadhruve10 (approved):

wideawakening:
thx @ravinadhruve10 let's see if this topic raises again

PR #482, opened 2024-02-27
feat(cloudauth): generic account component metadata

cgeers:
treat cloudauth account component metadata generic way
update cloudauth protobuf


sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -152,7 +153,7 @@ func resourceSysdigSecureCloudauthAccount() *schema.Resource {
 				Elem:     accountFeatures,
 			},
 			SchemaComponent: {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
```
cgeers, line 156: This change appears to remove the problem where component ordering surfaced spurious diff errors.
cc: @ravinadhruve10 @wideawakening

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -608,157 +498,86 @@ func featureToResourceData(features *cloudauth.AccountFeatures) []interface{} {
...
+	Gcp *internalServicePrincipalMetadata_GCP `json:"gcp,omitempty"`
 }
 
-func getResourceComponentsOrder(dataComponents interface{}) []string {
-	var dataComponentsOrder []string
-	for _, rc := range dataComponents.([]interface{}) {
-		resourceComponent := rc.(map[string]interface{})
-		dataComponentsOrder = append(dataComponentsOrder, resourceComponent[SchemaInstance].(string))
+func getComponentMetadataString(message protoreflect.ProtoMessage) string {
+	// marshal through protojson get correct snake case keys
```
ravinadhruve10, line 568: protojson would give us camelCase from cloudauth. And we re-marshal that using json to give us snake_case to match TF, correct?

ravinadhruve10 (approved):
Change looks good, thank you! One clarification question.

sysdig/resource_sysdig_secure_cloud_auth_account.go
cgeers: protojson can do either camel or snake, and invoke it for snake. the reason we're using both is because the key order that encoding/json uses some kind of key ordering that agrees with terraform marshalling.

we use protojson at all because it makes the code far cleaner.

PR #480, opened 2024-02-14
feat(cloud-host-scanner): enables cloud_account GCP WIF usage [SSPROD-35921]

wideawakening:
- ✅ upgrade proto
- 🎯 update module usage to handle WIFs
- 🎯 tests
- doc

PR #487, opened 2024-03-05
feat(onboarding,cloud-host-scanner): v2 enables cloud_account GCP WIF usage [SSPROD-35921] 

wideawakening:
duplicated from #480,  after resolving conflicts on master (to include #482).
had to add a small fix since tests were not passing 

--- 

- ✅ upgrade proto
- ✅  update module usage to handle WIFs
- ✅    tests
  - requires cloud-auth changes to be deployed to prod
  - clarifiation on https://github.com/sysdiglabs/terraform-provider-sysdig/pull/481
- ✅  NOOP. doc 
  - there is not much insight on component/feature by now, so not gonna add this scanning specific
-  ✅  do a real test on int01 with local provider


sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -441,3 +441,66 @@ func TestAccAWSSecureCloudAccountFCCSPM(t *testing.T) {
 		},
 	})
 }
+
+func XTestGCPAgentlesScanningOnboarding(t *testing.T) {
```
wideawakening, line 445: this test is not passing now, but on the original PR it did.
need some help here @cgeers  :)
any ideas? i'm guessing it has to do with the un/marshalling.. but not sure

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -555,10 +574,12 @@ func componentsToResourceData(components []*cloudauth.AccountComponent) []map[st
 
 // internal type redefintion for GCP service principals.
 // This exists because in terraform, the key is originally provided in the form of a base64 encoded json string
+
+// note; caution with order of fields, they have to go in alphabetical ASC so that the json marshalled on the tf read phase produces no drift https://github.com/golang/go/issues/27179
```
wideawakening, line 578: cc @cgeers just so you are aware of this.
for me it's ok to merge with this

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -387,26 +387,42 @@ func constructAccountComponents(data *schema.ResourceData) []*cloudauth.AccountC
...
-											Email:                      spGcp.Gcp.Email,
-										},
-									},
-								},
+							metadata.ServicePrincipalMetadata.GetGcp().Key = spGcpKey
+						}
+
+						if hasWifData {
+							metadata.ServicePrincipalMetadata.GetGcp().WorkloadIdentityFederation = &cloudauth.ServicePrincipalMetadata_GCP_WorkloadIdentityFederation{
+								PoolProviderId: spGcp.Gcp.WorkloadIdentityFederation.PoolProviderId,
```
ravinadhruve10, line 420: Not a blocker, but since we are updating this code, can we ensure passing all the fields within `GCP_WorkloadIdentityFederation` as received, rather than passing just one of them i.e `PoolProviderId` ?

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -441,3 +441,140 @@ func TestAccAWSSecureCloudAccountFCCSPM(t *testing.T) {
...
+		},
+		ProviderFactories: map[string]func() (*schema.Provider, error){
+			"sysdig": func() (*schema.Provider, error) {
+				return sysdig.Provider(), nil
+			},
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+		resource "sysdig_secure_cloud_auth_account" "gcp-agentless-scanning" {
```
ravinadhruve10, line 521: Just curious, did the alignment give you any problem here while running the tests?

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -387,26 +387,42 @@ func constructAccountComponents(data *schema.ResourceData) []*cloudauth.AccountC
 					if data.Get(SchemaCloudProviderType).(string) == cloudauth.Provider_PROVIDER_GCP.String() {
 						spGcp := &internalServicePrincipalMetadata{}
 						err = json.Unmarshal([]byte(value.(string)), spGcp)
-						if len(spGcp.Gcp.Key) >= 0 {
```
cgeers, line 390: `>=` is a legitimate, but the only bug here. Should be `>`, which I'd like to see be the only change within this section

cgeers (changes requested):
I've applied your test cases to `main`, and they consistently pass with after the `Email` order change and the `>=` -> `>` change in L390.  Can we please revert the GCP handling to the former structure?

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
wideawakening: both on that line and in general,  yes, the current tests strategy is a pain in the ass because it's space-sensitive.
i think that's something worth to investigate; on howto make it more lightweight, if possible, but did not check any options myself.

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -387,26 +387,42 @@ func constructAccountComponents(data *schema.ResourceData) []*cloudauth.AccountC
...
+						}
+
+						if hasWifData {
+							metadata.ServicePrincipalMetadata.GetGcp().WorkloadIdentityFederation = &cloudauth.ServicePrincipalMetadata_GCP_WorkloadIdentityFederation{
+								PoolProviderId: spGcp.Gcp.WorkloadIdentityFederation.PoolProviderId,
 							}
+							metadata.ServicePrincipalMetadata.GetGcp().Email = spGcp.Gcp.Email
 						}
+
+						component.Metadata = metadata
```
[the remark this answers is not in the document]
wideawakening: these changes were required due to the existing line  on
```
if len(spGcp.Gcp.Key) >= 0 {
```
in our new use-case there is no Key, so i I had to add some validation to discern the new use case, leaving us with two use cases
- key data is present > we populate key
- wif data is present (+email as dependency) > we populate wif

that's what the current code conveys. the `ifs` are not for validation, but for separating the two use cases.

i was careful not to be very disruptive with existing code, but if i did not reach my intention (may be my lack of go language expertise), feel free to modify it as required.  as long as the tests are agreed upon us, feel free to change the impl as you see it fit

sysdig/resource_sysdig_secure_cloud_auth_account.go
wideawakening: haha, yeah...  if you want to dig more into the issue feel free to do so, but let's agree on some criteria to accept this PR and unblock us;  you can work on digging more  anytime later.

i was there for couple of days and i'm ok letting it go... 
gotta say though, i'm not expert on go, neither the json libraries. maybe you do see some other option

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
ravinadhruve10: I believe its to do with TF files in general. If there is a func or parser to convert free form formatted string to TF acceptable format, that would make it lightweight.

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: This comment is a noop now, with the change block removed.

ravinadhruve10 (approved):

cgeers (approved):
Thanks @wideawakening looks good!

PR #494, opened 2024-03-25
Rules order in policy showing drift during apply

jacklongsd:

### DIFF
--- a/sysdig/resource_sysdig_secure_custom_policy.go
+++ b/sysdig/resource_sysdig_secure_custom_policy.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"reflect"
 	"strconv"
 	"time"
 
@@ -141,7 +140,7 @@ func customPolicyToResourceData(policy *v2.Policy, d *schema.ResourceData) {
 	}
 	fmt.Printf("Current rules: %+v", currentRules)
 	fmt.Printf("New rules: %+v", newRules)
-	areRulesSame := reflect.DeepEqual(currentRules, newRules)
+	areRulesSame := arePolicyRulesEquivalent(currentRules, newRules)
 	if !areRulesSame {
 		_ = d.Set("rules", newRules)
 	} else {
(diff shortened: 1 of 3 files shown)


sysdig/resource_sysdig_secure_custom_policy_test.go
```diff
@@ -25,13 +25,16 @@ func TestAccCustomPolicy(t *testing.T) {
...
-				Config: customPolicyWithName(rText()),
+				Config: customPolicyWithName(policy1),
 			},
 			{
 				ResourceName:      "sysdig_secure_custom_policy.sample",
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
+			{
+				Config: customPolicyWithRulesOrderChange(policy1),
```
ombellare, line 36: Nit: can we keep this as `rText()` for consistency as all the other tests already use it? Or you would need to change all tests to use your new variable

sysdig/resource_sysdig_secure_custom_policy.go
```diff
@@ -123,14 +123,66 @@ func customPolicyToResourceData(policy *v2.Policy, d *schema.ResourceData) {
...
 	}
-	_ = d.Set("rules", rules)
+	currentRules := []map[string]interface{}{}
+	for _, rule := range rules {
+		currentRules = append(currentRules, map[string]interface{}{
+			"name":    rule.Name,
+			"enabled": rule.Enabled,
+		})
+	}
+	fmt.Printf("Current rules: %+v", currentRules)
```
ombellare, line 141: Can we remove the Printfs added for debugging in this file please?

sysdig/resource_sysdig_secure_custom_policy.go
```diff
@@ -123,14 +123,66 @@ func customPolicyToResourceData(policy *v2.Policy, d *schema.ResourceData) {
...
+	fmt.Printf("New rules: %+v", newRules)
+	areRulesSame := arePolicyRulesEquivalent(currentRules, newRules)
+	if !areRulesSame {
```
ombellare, line 144: Nit: `if !arePolicyRulesEquivalent(currentRules, newRules)` to remove the extra variable

sysdig/resource_sysdig_secure_custom_policy.go
```diff
@@ -123,14 +123,66 @@ func customPolicyToResourceData(policy *v2.Policy, d *schema.ResourceData) {
...
+	currentRules := []map[string]interface{}{}
+	for _, rule := range rules {
+		currentRules = append(currentRules, map[string]interface{}{
+			"name":    rule.Name,
+			"enabled": rule.Enabled,
+		})
+	}
+	fmt.Printf("Current rules: %+v", currentRules)
+	fmt.Printf("New rules: %+v", newRules)
+	areRulesSame := arePolicyRulesEquivalent(currentRules, newRules)
```
ombellare, line 143: Can we use `reflect.DeepEqual` provided by Go for checking equality of maps? (https://techwasti.com/the-best-way-to-check-the-equality-of-two-maps-is-in-go). It would be tested for all edge cases and reduce the code that we need to maintain

ombellare (commented):
Suggested some improvements. Overall LGTM

sysdig/resource_sysdig_secure_custom_policy.go
jacklongsd: I tried using that at first, took me forever to realize it wasn't working. The problem is the data structure is a slice of maps, and when doing reflect.DeepEquals it cares about the order of elements which is exactly what I am trying to avoid.

sysdig/resource_sysdig_secure_custom_policy_test.go
jacklongsd: rText() generates a random string to make each policy different. The purpose of the test is to create a policy, modify it's rules order and not see drift on tf plan. Need the names to be the same to achieve this

ombellare (approved):
LGTM! Makes sense to get one more team approval before merging 👍

PR #504, opened 2024-04-24
feat(secure-onboarding) Adding AccountComponent Resource

ravinadhruve10:
Change summary:
----------------
- Adding new Account Component resource with schema and CRUD operations in parity with Cloudauth Account resource.
- Adding the respective client and support for new resource type.
- Refactored common parsing methods for conversion from/to component metadata.
- Added TF ACC tests for the new resource type.
- Added docs md for the new resource

### DIFF
--- a/sysdig/internal/client/v2/cloudauth.go
+++ b/sysdig/internal/client/v2/cloudauth.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
 const (
@@ -26,7 +27,7 @@ type CloudauthAccountSecureInterface interface {
 }
 
 func (client *Client) CreateCloudauthAccountSecure(ctx context.Context, cloudAccount *CloudauthAccountSecure) (*CloudauthAccountSecure, string, error) {
-	payload, err := client.marshalProto(cloudAccount)
+	payload, err := client.marshalCloudauthProto(cloudAccount)
 	if err != nil {
 		return nil, "", err
 	}
@@ -42,7 +43,8 @@ func (client *Client) CreateCloudauthAccountSecure(ctx context.Context, cloudAcc
 		return nil, errStatus, err
 	}
 
-	cloudauthAccount, err := client.unmarshalProto(response.Body)
+	cloudauthAccount := &CloudauthAccountSecure{}
+	err = client.unmarshalCloudauthProto(response.Body, cloudauthAccount)
 	if err != nil {
 		return nil, "", err
 	}
@@ -62,7 +64,8 @@ func (client *Client) GetCloudauthAccountSecure(ctx context.Context, accountID s
 		return nil, errStatus, err
 	}
 
-	cloudauthAccount, err := client.unmarshalProto(response.Body)
+	cloudauthAccount := &CloudauthAccountSecure{}
+	err = client.unmarshalCloudauthProto(response.Body, cloudauthAccount)
 	if err != nil {
 		return nil, "", err
 	}
@@ -84,7 +87,7 @@ func (client *Client) DeleteCloudauthAccountSecure(ctx context.Context, accountI
 
 func (client *Client) UpdateCloudauthAccountSecure(ctx context.Context, accountID string, cloudAccount *CloudauthAccountSecure) (
 	*CloudauthAccountSecure, string, error) {
-	payload, err := client.marshalProto(cloudAccount)
+	payload, err := client.marshalCloudauthProto(cloudAccount)
 	if err != nil {
 		return nil, "", err
 	}
@@ -100,7 +103,8 @@ func (client *Client) UpdateCloudauthAccountSecure(ctx context.Context, accountI
 		return nil, errStatus, err
 	}
 
-	cloudauthAccount, err := client.unmarshalProto(response.Body)
+	cloudauthAccount := &CloudauthAccountSecure{}
+	err = client.unmarshalCloudauthProto(response.Body, cloudauthAccount)
 	if err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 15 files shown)


sysdig/internal/client/v2/cloudauth.go
```diff
@@ -119,22 +123,20 @@ func (client *Client) getCloudauthAccountURL(accountID string, decrypt string) s
 	return fmt.Sprintf(getCloudauthAccountPath, client.config.url, accountID, decrypt)
 }
 
-// local function for protojson based marshal/unmarshal of cloudauthAccount proto
-func (client *Client) marshalProto(data *CloudauthAccountSecure) (io.Reader, error) {
-	payload, err := protojson.Marshal(data)
+// common func for protojson based marshal/unmarshal of any cloudauth proto
+func (client *Client) marshalCloudauthProto(message protoreflect.ProtoMessage) (io.Reader, error) {
```
cgeers, line 127: nit: using `protobuf.Message` a mere type alias of `protoreflect.ProtoMessage` is preferred (for some reason).

cgeers: nit: this function doesn't really have anything to with cloudauth now, it's a protojson Marshaller.

sysdig/internal/client/v2/model.go
```diff
@@ -605,6 +605,10 @@ type CloudauthAccountSecure struct {
 	cloudauth.CloudAccount
 }
 
+type CloudauthAccountComponentSecure struct {
+	cloudauth.AccountComponent
```
cgeers, line 609: Is there a purpose/benefit for the struct embedding here versus type alias versus neither?

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
```diff
@@ -0,0 +1,390 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(timeout),
+			Update: schema.DefaultTimeout(timeout),
+			Read:   schema.DefaultTimeout(timeout),
+			Delete: schema.DefaultTimeout(timeout),
+		},
+		Schema: map[string]*schema.Schema{
```
cgeers, line 38: this schema definition seems to be largely duplicative of the component schema embedded within the account. The only difference being the presence of the `account_id` field. Can these definitions be consolidated, and merely add the account_id when needed at this top level context?

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
```diff
@@ -97,7 +97,7 @@ resource "sysdig_secure_cloud_auth_account" "gcp_config_posture" {
 	}
 	secure_identity_entitlement {
 	  enabled    = true
-	  components = ["COMPONENT_SERVICE_PRINCIPAL/secure-posture"]
+	  components = ["COMPONENT_WEBHOOK_DATASOURCE/secure-runtime"]
```
cgeers, line 100: this is already fixed up on master

cgeers (approved):
I think this largely looks good. I understand there are some functional cases that need to be addressed, but largely, this approach seems sound. Leaving a ✅ here even though it's likely more changes will come (and a rebase is needed).

sysdig/resource_sysdig_secure_cloud_auth_account_test.go
ravinadhruve10: Yes, this PR got pushed before the fix. Rebased it.

sysdig/internal/client/v2/cloudauth.go
ravinadhruve10: Yes, since all the cloudauth related resources (account, org and now AccountComponent) require this marshaling/unmarshaling, I refactored the func in a common place and to accept _any_ proto message type. Let me check changing the type to `protobuf.Message`

sysdig/internal/client/v2/model.go
ravinadhruve10: Its mainly to
- keep the convention standards, and
- if in future we want to add and process any additional fields in the provider (not backend), the struct gives us the room to do so without any major changes.

sysdig/internal/client/v2/cloudauth.go
ravinadhruve10: Done.

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
ravinadhruve10: Discussed this offline on the reasoning and consolidated this in latest commit.

sysdig/internal/client/v2/model.go
cgeers: I'll interpret this as _not really_, but I was just curious. I concede there is a place for convention following even if it yields no practical benefit, however there's also quite a bit of embedding scattered throughout this code base, which can have some negative side effects in large code bases if left unchecked. Consider this me "checking" on it. 😄

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -21,6 +19,59 @@ import (
 	cloudauth "github.com/draios/terraform-provider-sysdig/sysdig/internal/client/v2/cloudauth/go"
 )
 
+/*
+declare common schemas used across resources here
+*/
+var (
+	accountComponents = &schema.Resource{
```
cgeers, line 26: nit: should this be singular?

cgeers (approved):
looks great @ravinadhruve10

sysdig/resource_sysdig_secure_cloud_auth_account.go
ravinadhruve10: actually yes, singular makes sense for SchemaComponent usage in both account and component resources. I'll fix this when I touch this part of code again.

PR #508, opened 2024-05-01
feat(secure-onboarding): Adding AccountFeature Resource

ravinadhruve10:
Change summary:
-------------------
- Adding new Account Feature resource with schema and CRUD operations
   in parity with Cloudauth Account and AccountComponent resource.
- Adding the respective client and support for new resource type.
- Added TF ACC tests for the new resource type.
- Added docs md for the new resource.

### DIFF
--- a/sysdig/internal/client/v2/cloudauth_account_feature.go
+++ b/sysdig/internal/client/v2/cloudauth_account_feature.go
@@ -7,8 +7,7 @@ import (
 )
 
 const (
-	cloudauthAccountFeaturePath = "%s/api/cloudauth/v1/accounts/%s/feature/%s" // PUT
-	// getCloudauthAccountPath        = "%s/api/cloudauth/v1/accounts/%s?decrypt=%s" // does GET require decryption?
+	cloudauthAccountFeaturePath = "%s/api/cloudauth/v1/accounts/%s/feature/%s" // GET, PUT, DEL
 )
 
 type CloudauthAccountFeatureSecureInterface interface {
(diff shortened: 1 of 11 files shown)


sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
```diff
@@ -0,0 +1,181 @@
...
+  })
+}
+
+resource "sysdig_secure_cloud_auth_account_feature" "azure_config_posture" {
+  account_id		         = sysdig_secure_cloud_auth_account.azure_sample.id
+  type                       = "FEATURE_SECURE_CONFIG_POSTURE"
+  enabled                    = true
+  components                 = ["COMPONENT_SERVICE_PRINCIPAL/secure-posture"]
+
+  depends_on = [ sysdig_secure_cloud_auth_account_component.azure_service_principal ]
```
ravinadhruve10, line 97: This is the part where it expresses dependency that a feature cannot be upserted on an account unless a valid related component is already there on the account. In other words, we want to make sure an account and a component resource are already created first, before creating feature.

Instead of explicit `depends_on`, I can use the type and instance in L75-76 to construct the `components` string value in L95 to express implicit dependency.

sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
```diff
@@ -0,0 +1,181 @@
...
+				ResourceName:      "sysdig_secure_cloud_auth_account.azure_sample",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func secureAzureWithServicePrincipalFeature(accountID string) string {
+	// to replicate user behavior, snippet creates an actual azure account and
```
ravinadhruve10, line 52: However, this is the context of the TF acceptance tests. We can discuss and clarify offline the contents of TF snippets from the perspective of a user.

ravinadhruve10:
A heads-up - the TF acceptance tests in other envs are known to fail, until the BE changes are not released in all envs.

sysdig/internal/client/v2/cloudauth_account_feature.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
+		return client.ErrorAndStatusFromResponse(response)
+	}
+	return "", nil
+}
+
+func (client *Client) UpdateCloudauthAccountFeatureSecure(ctx context.Context, accountID, featureType string, cloudAccountFeature *CloudauthAccountFeatureSecure) (
```
cgeers, line 80: nit: create and update are the same thing and can share an implementation.

cgeers (approved):
This looks good to me. Everything I'd expect to see. Nicely done.

haresh-suresh (approved):

sysdig/internal/client/v2/cloudauth_account_feature.go
ravinadhruve10: Good call, updated in next commit.

PR #510, opened 2024-05-08
fix(tests): Skip onprem unsupported policy types

jacalvo:
* Do not run cloud policy types tests with `tf_acc_onprem_secure` tag

* Use TypeSet instead of TypeList for rules

### DIFF
--- a/sysdig/resource_sysdig_secure_custom_policy_test.go
+++ b/sysdig/resource_sysdig_secure_custom_policy_test.go
@@ -82,11 +82,11 @@ resource "sysdig_secure_custom_policy" "sample" {
   runbook = "https://sysdig.com"
 
   rules {
-    name = sysdig_secure_rule_falco.terminal_shell.name
+    name = "Write below etc"
     enabled = true
   }
   rules {
-    name = "Write below etc"
+    name = sysdig_secure_rule_falco.terminal_shell.name
     enabled = true
   }
 
--- a/sysdig/resource_sysdig_secure_custom_policy_test.go
+++ b/sysdig/resource_sysdig_secure_custom_policy_test.go
@@ -10,12 +10,54 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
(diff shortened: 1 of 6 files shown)


sysdig/resource_sysdig_secure_custom_policy_test.go
```diff
@@ -10,59 +10,62 @@ import (
...
+		},
+		{
+			Config: customPoliciesWithKillAction(rText()),
+		},
+		{
+			Config: customPoliciesWithDisabledRules(rText()),
+		},
+	}
+
+	if !buildinfo.OnpremSecure {
```
rosenbloomb-sysdig, line 53: Love this method. I was thinking we would have to separate out these steps into new file. Great job

rosenbloomb-sysdig (approved):
Awesome!

papajulio (approved):

dbonf:
> Change rules order in customPolicyWithName() to fix TestAccCustomPolicy

don't you think this can happen in a real life scenario? if order is not important maybe you want to use TypeSet instead of TypeList [here](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/f6815754e4e0d44fe4c35397a83964363f5ff258/sysdig/resource_sysdig_secure_custom_policy.go#L54)

jacalvo:
> > Change rules order in customPolicyWithName() to fix TestAccCustomPolicy
> 
> don't you think this can happen in a real life scenario? if order is not important maybe you want to use TypeSet instead of TypeList [here](https://github.com/sysdiglabs/terraform-provider-sysdig/blob/f6815754e4e0d44fe4c35397a83964363f5ff258/sysdig/resource_sysdig_secure_custom_policy.go#L54)

@dbonf I've pushed 2d8dbed changing to `TypeSet` and reverting the no longer necessary rule order change in the test. Please let me know if it's ok to edit the description and go ahead with the merge or better to split in two different PRs as this change may be a bit out of scope of the original purpose.

rosenbloomb-sysdig (approved):
Great job!

jcelaya (approved):
:heart:

storres1912 (approved):
Thank you @jacalvo!

kmvachhani (approved):

miguelgordo (approved):

PR #512, opened 2024-05-21
feat(onboarding): data source tenant external id

cgeers:
* adds external id as data source for use by sysdig tenants


sysdig/data_source_sysdig_secure_tenant_external_id.go
```diff
@@ -0,0 +1,49 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+// Retrieves the information of a resource form the file and loads it in Terraform
+func dataSourceSysdigSecureTenantExternalIDRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := getSecureCloudAccountClient(meta.(SysdigClients))
```
ravinadhruve10, line 32: This makes these APIs via the cloud_account (i.e cloudsec client). If we plan to deprecate the client, we will need to move these API calls to the cloudauth based client.

sysdig/internal/client/v2/cloud_account.go
```diff
@@ -11,7 +11,8 @@ const (
 	cloudAccountsWithExternalIDPath = "%s/api/cloud/v2/accounts?includeExternalID=true&upsert=true"
 	cloudAccountPath                = "%s/api/cloud/v2/accounts/%s"
 	cloudAccountWithExternalIDPath  = "%s/api/cloud/v2/accounts/%s?includeExternalID=true"
-	trustedCloudIdentityPath        = "%s/api/cloud/v2/%s/trustedIdentity"
+	onboardingTrustedIdentityPath   = "%s/api/secure/onboarding/v2/trustedIdentity?provider=%s"
```
ravinadhruve10, line 14: The actual source value will be the same, so this won't break any backward compatibility right?

ravinadhruve10 (commented):
Couple of clarifying questions, overall change lgtm.

ravinadhruve10:
We will need a docs change as well in this PR, since new data sources are being added.

sysdig/internal/client/v2/cloud_account.go
cgeers: yes, the value is the same.

sysdig/data_source_sysdig_secure_tenant_external_id.go
cgeers: good callout. I hadn't noticed this.

cgeers: it's a good callout, but these "clients" aren't actually different things. It's just a means to organize code along organizational boundaries.  Let me see if I can organize this in a better way.

cgeers: I've created an onboarding "client", that consolidates things a bit.

ravinadhruve10 (approved):
Changes look good. However, CI is failing with below :-
- [test fails](https://github.com/sysdiglabs/terraform-provider-sysdig/actions/runs/9275708453/job/25521093291?pr=512#step:4:987) in the CI pipeline
- we would need docs change as well since new data sources are being added
- lint fix
can we please check these?

ravinadhruve10 (approved):

dbonf (approved):

PR #533, opened 2024-07-12
fix(cloudauth): remove 404 error swallow

cgeers:
These errors should not be swallowed. they are legitimate and should be bubbled up. Leaving the 404 swallow for the delete case since it also represents success on a delete operation


cgeers:
@ravinadhruve10 within the constraints of this older terraform testing framework, I'm not certain this best approach to apply tests for this behavior. any thoughts?

ravinadhruve10:
> @ravinadhruve10 within the constraints of this older terraform testing framework, I'm not certain this best approach to apply tests for this behavior. any thoughts?

Sorry, not sure I follow. Are we asking if the current ACC tests we have are testing all the possible behaviors or not? (the ones in question particularly)

ravinadhruve10 (approved):
Since the current ACC tests are passing without this, it is no longer required. LGTM.

PR #534, opened 2024-07-23
feat(ip filter) Add ip filter resource for both monitor and secure

vojindj:


sysdig/internal/client/v2/allowed_ip_ranges.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (client *Client) DeleteAllowedIpRange(ctx context.Context, id int) error {
+	response, err := client.requester.Request(ctx, http.MethodDelete, client.GetAllowedIpRangeURL(id), nil)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK && response.StatusCode != http.StatusNotFound {
```
filiptubic, line 101: Why having `response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK`? I assume server responds only with StatusOK?

sysdig/internal/client/v2/allowed_ip_ranges.go
```diff
@@ -0,0 +1,114 @@
...
+var AllowedIpRangeNotFound = errors.New("IP range not found")
+
+const (
+	AllowedIPRangesPath = "%s/platform/v1/allowed-ip-ranges"
+	AllowedIPRangePath  = "%s/platform/v1/allowed-ip-ranges/%d"
+)
+
+type AllowedIPRangesInterface interface {
+	Base
+	GetAllowedIpRangeById(ctx context.Context, id int) (*AllowedIpRange, error)
```
filiptubic, line 19: In go we should name this `GetAllowedIPRangeByID` this comment applies for all other methods

sysdig/internal/client/v2/model.go
```diff
@@ -174,6 +174,14 @@ type GroupMappingConfig struct {
 	DifferentTeamSameRoleStrategy string `json:"differentRolesSameTeamStrategy"`
 }
 
+type AllowedIpRange struct {
+	ID          int    `json:"id,omitempty"`
+	IpRange     string `json:"ipRange"`
```
filiptubic, line 179: IPRange

sysdig/internal/client/v2/allowed_ip_ranges.go
vojindj: Actually, StatusNoContent is returned, so I removed check for StatusOK.

sysdig/internal/client/v2/allowed_ip_ranges.go
vojindj: updated

sysdig/internal/client/v2/model.go
vojindj: updated

dbonf (approved):
LGTM, I get that we are not able to test the write path of `sysdig_ip_filtering_settings` in integration tests otherwise we cut ourself out of the environment, we can live with that.

Shadow649 (approved):

PR #539, opened 2024-07-30
feat(cloudauth): component version

cgeers:


haresh-suresh (approved):

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -33,6 +33,10 @@ var (
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			SchemaVersion: {
```
ravinadhruve10, line 36: @cgeers I believe we will need to update the component resource as well with this change.

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -484,6 +490,7 @@ func componentsToResourceData(components []*cloudauth.AccountComponent) []map[st
 		resourceData := map[string]interface{}{}
 		resourceData[SchemaType] = component.GetType().String()
 		resourceData[SchemaInstance] = component.GetInstance()
+		resourceData[SchemaVersion] = component.GetVersion()
```
ravinadhruve10, line 493: Shall we update the ACC tests as well once BE PR is merged?

cgeers: I don't think this actually depends on new BE functionality. We can add the field in tests now.

cgeers: done

sysdig/resource_sysdig_secure_cloud_auth_account.go
cgeers: good call. Yes, this already serves the immediate intended purpose, but there's no reason to leave this half implemented. I've added added the conversion steps to the component resourced as well.

PR #541, opened 2024-08-02
feat(onboarding): cloud ingestion assets

cgeers:


sysdig/data_source_sysdig_secure_onboarding.go
```diff
@@ -282,6 +282,72 @@ func dataSourceSysdigSecureAgentlessScanningAssetsRead(ctx context.Context, d *s
...
+				},
+			},
+			"gcp_routing_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"gcp_metadata": {
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
```
nkraemer-sysdig, line 310: Any reason we use a map for AWS, but individual strings for GCP?

ravinadhruve10: So we get a map of values from the API, both for AWS and GCP. Original change was using a map for both.

In GCP case however, the map itself is a mix of string and another map. To be able to read GCP's values, we would have to do nested map in the schema then. That was making it complicated to parse and do schema conversions. Hence this separation of routing key and metadata looks cleaner and TF can directly read them.

We can change AWS to use string too for the ARN. But incase we add more ingestion related assets in future, this would need to change.

haresh-suresh (approved):

sysdig/data_source_sysdig_secure_onboarding.go
nkraemer-sysdig: Makes sense, thanks for the explanation!

nkraemer-sysdig (approved):

sysdig/data_source_sysdig_secure_onboarding.go
ravinadhruve10: Ofcourse, np!

PR #547, opened 2024-08-19
chore(docs): Add Serverless Agent examples

pgcrooks-sysdig:
Add Serverless Agent example deployments. These will be referenced by updated public Docs.

Ref: DOC-4176

### DIFF
--- a/examples/serverless-agent/fargate/workload/instrumented_load.tf
+++ b/examples/serverless-agent/fargate/workload/instrumented_load.tf
@@ -1,0 +1,145 @@
+data "sysdig_fargate_workload_agent" "containers_instrumented" {
+  container_definitions = jsonencode([
+    {
+      "name" : "event-gen-1",
+      "image" : "falcosecurity/event-generator",
+      "command" : ["run", "syscall", "--all", "--loop"],
+      "logConfiguration" : {
+        "logDriver" : "awslogs",
+        "options" : {
+          "awslogs-group" : aws_cloudwatch_log_group.instrumented_logs.name,
+          "awslogs-region" : var.region,
+          "awslogs-stream-prefix" : "task"
+        },
+      }
+    },
+    {
+      "name" : "event-gen-2",
+      "image" : "falcosecurity/event-generator",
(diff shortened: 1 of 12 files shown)


examples/serverless-agent/fargate/workload/instrumented_load.tf
```diff
@@ -0,0 +1,145 @@
...
+}
+
+resource "aws_security_group_rule" "orchestrator_agent_ingress_rule" {
+  type              = "ingress"
+  protocol          = "tcp"
+  from_port         = 0
+  to_port           = 0
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.security_group.id
+}
```
gnosek, line 136: I'm assuming this only gives access to the security group, right? not to the whole world

examples/serverless-agent/fargate/workload/providers.tf
```diff
@@ -0,0 +1,4 @@
+provider "aws" {
+  region = var.region
+  profile = var.profile
+}
```
francesco-racciatti, line 4: Shouldn't we add the sysdig provider here?

francesco-racciatti: oh, nvm, it's in versions.tf

francesco-racciatti (approved):
lgtm

marojor (approved):
LGTM

examples/serverless-agent/fargate/workload/instrumented_load.tf
pgcrooks-sysdig: This has long caused pain, since the group is completely open. Ideally customers would limit this to _just_ the collector range, but this example doesn't even attempt to be secure :-)

examples/serverless-agent/fargate/workload/providers.tf
pgcrooks-sysdig: Happy to combine / rearrange. There's a dozen ways to lay out these projects.

PR #548, opened 2024-08-26
feat(cloudauth) SSPROD-38727: updating cloudauth pb.go file to proxy …

haresh-suresh:
…regions in EB


cgeers (approved):

cgeers:
looks like a linting error of some sort.  Since this file is generated, seems reasonable to omit this file from this kind of check.

PR #554, opened 2024-09-23
[SSPROD-46998] Adding agentless azure sysdig_secure_trusted_azure_app

miguelpais:
The sysdig azure application needs to be added to the terraform provider so that it can be used by the dependent terraform modules.

Se associated PR:
https://github.com/draios/secure-backend/pull/36116


cgeers (changes requested):
acceptance tests are failing. azure_app name is wrong

miguelpais:
> acceptance tests are failing. azure_app name is wrong

You are completely right @cgeers , thanks, updated it!

cgeers (approved):
👍 but looks like this needs a rebase

PR #559, opened 2024-10-04
feat(cloudauth): support accounts with provider partition field

ravinadhruve10:
Change summary:
-----------------
- Add support for onboarding cloudauth accounts with regulatory framework type
- Updated cloudauth proto pb.go


sysdig/resource_sysdig_secure_cloud_auth_account_component_test.go
```diff
@@ -59,6 +59,7 @@ resource "sysdig_secure_cloud_auth_account" "azure_sample" {
   enabled            = true
   provider_tenant_id = "%s"
   provider_alias     = "some-alias"
+  regulatory_framework = "REGULATORY_FRAMEWORK_UNSPECIFIED"
```
ravinadhruve10, line 62: I don't believe this is necessary. Not specifying the regulatory_framework here will make protobuf automatically set it to REGULATORY_FRAMEWORK_UNSPECIFIED. Basically the meaning of this enum value itself is that the field was not provided/populated in the request.

The ACC tests are failing because of BE changes not available in prod envs. Once they start passing, without this change also they should pass IMO.

jose-pablo-camacho: Yeah, we discussed this so I'll get rid off these

website/docs/r/secure_cloud_auth_account.md
```diff
@@ -42,4 +45,6 @@ In addition to all arguments above, the following attributes are exported:
 
 * `id` - (Computed) The ID of the cloud account.
 
-* `organization_id` - (Computed) The ID of the organization, if the cloud account is part of any organization.
\ No newline at end of file
+* `organization_id` - (Computed) The ID of the organization, if the cloud account is part of any organization.
+
+* `regulatory_framework` - (Computed) The type of the regulatory framework for the account.
```
cgeers, line 50: this is not a computed value

cgeers (changes requested):

website/docs/r/secure_cloud_auth_account.md
jose-pablo-camacho: updated the md file

website/docs/r/secure_cloud_auth_account.md
```diff
@@ -34,6 +34,9 @@ resource "sysdig_secure_cloud_auth_account" "sample" {
 
 * `component` - (Optional) The component configuration to enable on this cloud account. There can be multiple component blocks for a feature, one for each component to be enabled.
 
+* `regulatory_framework` - (Optional - AWS installs only) The type of Regulatory Framework for cloud account. Currently supported options are `REGULATORY_FRAMEWORK_UNSPECIFIED` and `RegulatoryFramework_REGULATORY_FRAMEWORK_US_FEDRAMP`.
+
```
ravinadhruve10, line 38: nit: can we keep the text consistent - and `REGULATORY_FRAMEWORK_US_FEDRAMP` ?

ravinadhruve10 (commented):
Lgtm, one minor comment.

website/docs/r/secure_cloud_auth_account.md
jose-pablo-camacho: done

haresh-suresh (approved):

sysdig/resource_sysdig_secure_cloud_auth_account.go
```diff
@@ -586,5 +591,13 @@ func cloudauthAccountToResourceData(data *schema.ResourceData, cloudAccount *v2.
 		}
 	}
 
+	// TODO: add an acc test with no values, with fedramp values and with ""
+	if !(cloudAccount.ProviderPartition.String() == cloudauth.ProviderPartition_name[0]) {
+		err = data.Set(SchemaProviderPartition, cloudAccount.ProviderPartition.String())
```
ravinadhruve10, line 596: is there a specific reason for choosing cloudauth.ProviderPartition_name[0] for comparing instead of the string directly?

ravinadhruve10 (commented):
one comment, rest LGTM.

sysdig/resource_sysdig_secure_cloud_auth_account.go
jose-pablo-camacho: Actually not, updated it to use the related const

cgeers (approved):

ravinadhruve10 (commented):
Since I was the PR author I can't approve it. But LGTM.

haresh-suresh (approved):

PR #587, opened 2025-01-08
docs(serverless): the workload agent no longer requires the orchestrator

francesco-racciatti:
This PR updates the documentation of the fargate workload agent data source. The orchestrator agent is no longer required.


pgcrooks-sysdig (approved):
LGTM

PR #591, opened 2025-01-10
feat(organization): add async option to organizations endpoint to support onboarding of large organizations

maratsal:
This PR is to address organizations onboarding with thousands of accounts, where previous API endpoint would timeout. `?async=true` option was introduced to API, but terraform provider update was missing.

This code has been tested with 900 accounts organizaton onboarding and all worked fine.


nkraemer-sysdig:
Its worth noting that this change will affect all cloud providers (not just AWS). I'm checking with the team to make sure there are no concerns there

maratsal:
I tested those manually and it was working fine. Also added async option to another endpoint as well (needed for PUT calls), can be done in more elegant way by concatenating URL in place only for 2 function: CreateOrganizationSecure and UpdateOrganizationSecure. But this way works too.

 And finally I was thinking that we likely can implement this by using async option on the api endpoint by default.

cgeers (changes requested):
This runs counter to the current plans in this area, I do not support merging this.

maratsal:
@cgeers which part you don't like? Should i redo it and introduce new option to set async parameter?

cgeers:
@maratsal Here are the problems with this change:

* the API behavior you're relying on planned to be removed in short order
* the change here is broad and sweeping and affects things far outside your intended scope
* if merged as is, QA tests will begin failing and it'll be up to someone to fix all of that

I think the best option here is to close this PR and let the team deliver the feature that supersedes all these workarounds. If that's not acceptable for you, instead of changing the default for or all organizations, make a more targeted change.  If you were to control this query param with an _undocumented_ environment variable, for example, while leaving the default behavior as is, that would be acceptable to merge, IMO.

maratsal:
@cgeers sounds good, I pushed changes to the code to set async=true parameter based on env variable only. Please let me know if this approach will work.

cgeers (approved):

maratsal:
closing this PR in favour of https://github.com/sysdiglabs/terraform-provider-sysdig/pull/594 to be able to finish tests.

PR #594, opened 2025-01-17
feat(organization): add hidden async option to organizations endpoint to support onboarding of large organizations

maratsal:
This PR is to address organizations onboarding with thousands of accounts, where previous API endpoint would timeout. ?async=true option was introduced to API, but terraform provider update was missing.

This code has been tested with 900 accounts organizaton onboarding and all worked fine.

replacing PR https://github.com/sysdiglabs/terraform-provider-sysdig/pull/591 to be able to finish tests.


maratsal:
@cgeers can you review this one please. tests here are passing now.

cgeers (approved):

PR #595, opened 2025-01-20
chore: update codeowners with new owners

tembleking:


tembleking:
Closing, since we don't need yet to be codeowners and we need to change an internal policy at sysdig.

PR #605, opened 2025-02-27
SSPROD-54060 - organization: add root org id

jose-pablo-camacho:
xref: https://sysdig.atlassian.net/browse/SSPROD-54060


sysdig/resource_sysdig_secure_organization.go
```diff
@@ -73,6 +73,11 @@ func resourceSysdigSecureOrganization() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			SchemaOrganizationRootID: {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
+			},
```
ravinadhruve10, line 80: nit: if we make this a required field, it will break backwards compatibility right?

ravinadhruve10 (commented):
Overall LGTM, good to also add unit tests.

sysdig/resource_sysdig_secure_organization.go
jose-pablo-camacho: Yes, that's why I added it as `optional: true` to ensure it will not break backwards compatibility

ravinadhruve10 (approved):
nit: could you also please update the docs for org object?

ravinadhruve10 (approved):

PR #609, opened 2025-03-11
feat(zone): add Zone resource for Secure

vojindj:


sysdig/internal/client/v2/zones.go
```diff
@@ -0,0 +1,121 @@
+package v2
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+const (
+	ZonesPath = "%s/platform/v1/zones"
```
luzeno, line 10: Not sure if we prefer to call them `PlatformZones` given that we have called the legacy ones `PostureZones`

luzeno (commented):
Looks good to me, but it would be better to get some more experienced review

sysdig/internal/client/v2/zones.go
vojindj: done

sysdig/resource_sysdig_secure_zone.go
```diff
@@ -0,0 +1,216 @@
...
+			SchemaLastModifiedBy: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			SchemaLastUpdated: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			SchemaScopesKey: {
+				Optional: true,
```
DagDigg, line 50: Sorry, I don't know terraform well.
Is this resource related to the new `platform` zones?
Because if it is, `scopes` are mandatory, and they can be 1 or more.
On [L:52](https://github.com/sysdiglabs/terraform-provider-sysdig/pull/609/files#diff-c64024666fc07c683741ac0c050fe570ea1fcbf390a2a12310acd8cc2e883c5aR52) I see `MaxItems:1` .
Also scopes ids are missing 🤔 
This is an example response from staging:
```json
// ....
"scopes": [
    {
        "id": 439653,
        "rules": "",
        "targetType": "azure"
    },
    {
        "id": 439654,
        "rules": "",
        "targetType": "kubernetes"
    },
    {
        "id": 439655,
        "rules": "",
        "targetType": "host"
    }
]
```

vojindj: thanks for checking! yes, this resource is for creating new `platform` zones, but since it's only for secure currently it is named like other secure resources.
I changed `scopes` to be required field and also I added `id` as a field for `scope`... nice catch! 👍 
MaxItems:1 is for "scopes" field - we can have only one of those. MinItems: 1 is set for "scope", because we need at least one of them.

vojindj: I removed `scopes` to simplify the schema

filiptubic (approved):

PR #613, opened 2025-03-19
docs(issue 612): updated docs to reflect potential 403 error

bashbang:
note 403 error when when not setting option share_with_current_team in monitor_notification_channels for users with lower permissions than 'admin'

Also cleaned tailing whitespace in updated files.


dbonf:
thanks for this contribution! would you mind also updating the docs for the `sysdig_secure_notification_channel_*` as well, for the sake of consistency?

bashbang:
> thanks for this contribution! would you mind also updating the docs for the `sysdig_secure_notification_channel_*` resurces as well, for the sake of consistency?

the secure notification channels don't have a share_with_current_team option.

dbonf (approved):

PR #631, opened 2025-04-26
chore(notification-channels): Make Slack webhook URL fields sensitive

akahenry:
Slack webhook URLs are sensitive according to Slack documentation (https://api.slack.com/messaging/webhooks):

> Keep it secret, keep it safe. Your webhook URL contains a secret. Don't share it online, including via public version control repositories. Slack actively searches out and revokes leaked secrets.

This PR makes them sensitive in Terraform as well so they don't get sent to logs and unexpectedly exposed.


dbonf (approved):
thanks for this contribution, it makes totally sense, so, lgtm

PR #638, opened 2025-05-19
feat(cloudlogs): bump protorepo to support new fields

lorenzo-merici:


sysdig/internal/client/v2/cloudauth/go/cloud_account.pb.go
```diff
@@ -1,10 +1,10 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.32.0
-// 	protoc        v4.25.3
+// 	protoc-gen-go v1.35.2
+// 	protoc        v5.29.3
 // source: cloud_account.proto
 
-package draiosproto
+package cloudauthproto
```
matteopasa, line 7: draiosproto I think it should not change

matteopasa (approved):

PR #656, opened 2025-07-14
feat(secure-policies): support new fields in drift policy 

ombellare:


jbainbridgesysdig (dismissed):

daniel-almeida (dismissed):

sysdig/resource_sysdig_secure_drift_policy_test.go
```diff
@@ -67,9 +67,6 @@ resource "sysdig_secure_drift_policy" "sample" {
     prohibited_binaries {
       items = ["/usr/bin/curl"]
     }
-	process_based_exceptions {
```
tembleking, line 70: Please do not modify previous tests, as they ensure that existing code that customers may be using is not going to break.

Add a new test case with the new sysdig_secure_drift_policy.sample with the new fields you added to the resource.

sysdig/resource_sysdig_secure_drift_policy_test.go
```diff
@@ -138,16 +139,17 @@ resource "sysdig_secure_drift_policy" "sample" {
     description = "Test Drift Rule Description"
 
     enabled = true
+    use_regex = true
```
tembleking, line 142: Same here, do not add it to a previous existing resource tested. Add a new test case for the `use_regex = true`.

jbainbridgesysdig (approved):

PR #657, opened 2025-07-14
feat(secure-policies): support new and fix existing fields in malware policy

ombellare:


ombellare:
Hi @airadier , please hold off on reviewing this PR for now since there is a schema change that might need to be communicated to customers before proceeding to merge the fix. 
I have added a short description of the change so feel free to provide any early feedback if you would like!

jbainbridgesysdig (approved):

PR #671, opened 2025-10-15
feat!: deprecate cloudsec resources

jose-pablo-camacho:
This pull request deprecates the resource responsible for making `cloudsec` calls in the Sysdig Terraform provider. 

Specifically, it removes the `sysdig_secure_cloud_account` resource and its related implementation, tests, and documentation. The changes include deleting the resource file, its test file, and the corresponding documentation, as well as cleaning up references in the provider’s registration and internal client code. This is part of a broader effort to phase out legacy `cloudsec` resource management from the provider.

### DIFF
--- a/sysdig/internal/client/v2/sysdig.go
+++ b/sysdig/internal/client/v2/sysdig.go
@@ -39,7 +39,6 @@ type SysdigSecure interface {
 	SysdigCommon
 	SecureCommon
 
-	CloudAccountSecureInterface
 	CloudauthAccountComponentSecureInterface
 	CloudauthAccountFeatureSecureInterface
 	CloudauthAccountSecureInterface


ivanbesinovic-sysdig (approved):

ravinadhruve10 (approved):

sysdig/resource_sysdig_secure_cloud_account.go
tembleking: We cannot just remove resources from the provider, because those may be in use by some customer. First you need to mark it as deprecated pointing to a replacement and then ensure no one else is using it.

tembleking: You can use the field `DeprecationMessage` within `schema.Resource` like:

```go
DeprecationMessage: "This resource is deprecated. Please use X resource instead.",
```

That will mark the resource as deprecated.

tembleking (approved):

PR #677, opened 2025-11-05
feat(vulnerability-policy): add admission control stage

tembleking:
This PR adds support for the "admission_control" stage in vulnerability policies.

This includes the addition of two new configurable fields within the `configuration` block for the "admission_control" stage:

- `failure_action`: Defines the action to take when a policy fails (e.g., "reject", "warn").
- `unknown_image_action`: Defines the action to take when an image is unknown (e.g., "reject", "rejectAndScan", "warn").


tembleking:
Blocked until the API is able to ingest the `admission_control` stage name.
```
Error: request body has an error: doesn't match schema #/components/schemas/CreatePolicyRequest: Error at "/stages/2/name": value is not one of the allowed values ["runtime","pipeline","registry"]
```

PR #688, opened 2026-01-28
feat(sso): add sysdig_sso_openid and sysdig_sso_saml resources

tembleking:
## Summary

- Add `sysdig_sso_openid` resource for OpenID Connect SSO configuration
- Add `sysdig_sso_saml` resource for SAML SSO configuration

Both resources use the `/platform/v1/sso-settings/` endpoint with different `type` discriminators.

### sysdig_sso_openid features:
- Issuer URL with optional metadata discovery
- Client ID/Secret authentication
- Manual metadata configuration when discovery is disabled
- Additional OAuth scopes support

### sysdig_sso_saml features:
- Metadata via URL or inline XML (mutually exclusive)
- Email parameter mapping
- Security settings (signature validation, signed assertions, destination verification, encryption)

### Common SSO features (both resources):
- Product selection (monitor/secure)
- Group mapping configuration
- Single logout support
- Auto user creation on login
- Optimistic locking via version field

## Test plan

- [ ] Run acceptance tests for OpenID: `go test ./sysdig -v -run TestAccSSOOpenID -tags=tf_acc_sysdig_secure -timeout 120m`
- [ ] Run acceptance tests for SAML: `go test ./sysdig -v -run TestAccSSOSaml -tags=tf_acc_sysdig_secure -timeout 120m`
- [ ] Verify import functionality for both resources
- [ ] Test update operations with version-based optimistic locking

### DIFF
--- a/sysdig/resource_sysdig_sso_openid.go
+++ b/sysdig/resource_sysdig_sso_openid.go
@@ -2,6 +2,7 @@ package sysdig
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"time"
 
@@ -90,7 +91,7 @@ func resourceSysdigSSOOpenID() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				ForceNew:    true,
-				Description: "A custom name for this SSO integration (cannot be changed after creation)",
+				Description: "A name to distinguish different SSO integrations (cannot be changed after creation)",
 			},
 
 			// OpenID specific optional fields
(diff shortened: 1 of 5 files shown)


sysdig/resource_sysdig_sso_openid.go
```diff
@@ -0,0 +1,421 @@
...
+			"group_mapping_attribute_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "groups",
+				Description: "The attribute name for group mapping",
+			},
+			"integration_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
dbonf: it is legal to change the integration name, is there any reason for the `ForceNew` here?

sysdig/resource_sysdig_sso_openid.go
```diff
@@ -0,0 +1,421 @@
...
+			},
+
+			// OpenID specific optional fields
+			"is_metadata_discovery_enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether to use automatic metadata discovery from the issuer URL",
+			},
+			"group_attribute_name": {
```
dbonf: i think we have an issue with the current public api, we should ignore this `group_attribute_name`, and only rely on `groupMappingAttributeName` at root level

PR #705, opened 2026-02-12
fix(ci): use unique names in macro append test to prevent flakiness

tembleking:
The `TestAccMacro` test Step 5 (`macroAppendToDefault`) hardcodes the macro name `"container"` — a built-in Sysdig macro. When multiple CI runs execute concurrently against the same Sysdig environment, they collide with the error: *"The field 'name' must not be the same as another Secure UI macro"*.

This is the same class of bug fixed in #697 for managed rulesets and custom policies.

The fix replaces the hardcoded built-in macro reference with a self-contained test that creates its own base macro with a unique `terraform_test_` prefixed name, then appends to it. Also documents the "avoid name collisions" pattern in AGENTS.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -333,6 +333,14 @@ func TestAccResourceXxx(t *testing.T) {
 - `randomText(len)` - Generate random strings for unique names
 - `sysdigOrIBMMonitorPreCheck(t)` - Check for either credential type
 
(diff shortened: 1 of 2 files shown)


alecron (approved):

tembleking:
Closing, this is unneeded because the issue was actually in the infra we were testing with.

PR #726, opened 2026-04-16
feat(secure): add sysdig_secure_zone_posture_policy_assignment resource

tembleking:
New resource `sysdig_secure_zone_posture_policy_assignment` that manages the association between a `sysdig_secure_zone` and a set of posture policy IDs via the CSPM API (`/api/cspm/v1/zones/{zoneId}/policies`).

This is a standalone resource (not embedded in `sysdig_secure_zone`) because the APIs have different lifecycles and it allows assigning policies to system zones the user didn't create. The resource ID is the zone ID (1:1 relationship), and `zone_id` is ForceNew since changing it means a different resource.

Depends on backend routing enablement: `draios/sysdigcloud-harness-cd#75479`

### DIFF
--- a/.github/workflows/ci-provider-docs.yaml
+++ b/.github/workflows/ci-provider-docs.yaml
@@ -20,36 +20,10 @@ jobs:
       with:
         go-version-file: go.mod
 
-    - name: go build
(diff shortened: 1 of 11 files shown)


mateobur (approved):

alecron (approved):
LGTM

PR #753, opened 2026-09-10
Use TypeSet for unordered organization collection attributes

ivanbesinovic-sysdig:
The include/exclude collections on `sysdig_secure_organization` were `TypeList`, so element order was part of the diff and every plan where the backend returned a different order showed a spurious in-place update — which sends a full org PUT and re-runs the org scrape on the backend. Verified live on AWS and Azure org onboarding: the released provider plans an update on a plain refresh, this build reports no changes.

Second commit: `account_id`, `type` and `instance` on `sysdig_secure_cloud_auth_account_component` are now `ForceNew`. They form the component's API path, and the update path looked the component up by its *new* identity, got a 404 and returned nil — so a rename silently wrote new state while the backend kept the old component. The guard meant to reject this was unreachable for the same reason and is removed.

Note for release notes: indexing the organization attributes (e.g. `excluded_cloud_accounts[0]`) no longer works, since sets are not indexable.

### DIFF
--- a/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
@@ -160,9 +160,6 @@ func resourceSysdigSecureCloudauthAccountComponentDelete(ctx context.Context, da
 	return nil
 }
 
-/*
-This function validates and restricts any fields not allowed to be updated during resource updates.
-*/
 func cloudauthAccountComponentFromResourceData(data *schema.ResourceData) *v2.CloudauthAccountComponentSecure {
 	cloudAccountComponent := &v2.CloudauthAccountComponentSecure{
 		AccountComponent: cloudauth.AccountComponent{
--- a/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
@@ -99,6 +99,7 @@ func resourceSysdigSecureCloudauthAccountComponentRead(ctx context.Context, data
 		ctx, data.Get(SchemaAccountID).(string), data.Get(SchemaType).(string), data.Get(SchemaInstance).(string))
 	if err != nil {
 		if strings.Contains(errStatus, "404") {
+			data.SetId("")
 			return nil
 		}
 		return diag.Errorf("Error reading resource: %s %s", errStatus, err)
@@ -119,21 +120,13 @@ func resourceSysdigSecureCloudauthAccountComponentUpdate(ctx context.Context, da
 	}
 
 	accountID := data.Get(SchemaAccountID).(string)
-	_, errStatus, err := client.GetCloudauthAccountComponentSecure(
-		ctx, accountID, data.Get(SchemaType).(string), data.Get(SchemaInstance).(string))
-	if err != nil {
-		if strings.Contains(errStatus, "404") {
-			return nil
-		}
-		return diag.Errorf("Error reading resource: %s %s", errStatus, err)
-	}
-
 	newCloudAccountComponent := cloudauthAccountComponentFromResourceData(data)
 
-	_, errStatus, err = client.UpdateCloudauthAccountComponentSecure(
+	_, errStatus, err := client.UpdateCloudauthAccountComponentSecure(
 		ctx, accountID, data.Get(SchemaType).(string), data.Get(SchemaInstance).(string), newCloudAccountComponent)
 	if err != nil {
 		if strings.Contains(errStatus, "404") {
+			data.SetId("")
 			return nil
 		}
 		return diag.Errorf("Error updating resource: %s %s", errStatus, err)
--- a/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account_component.go
@@ -6,7 +6,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"maps"
 	"strings"
 	"time"
@@ -46,10 +45,21 @@ func getAccountComponentSchema() map[string]*schema.Schema {
 		SchemaAccountID: {
 			Type:     schema.TypeString,
 			Required: true,
+			ForceNew: true,
 		},
 	}
 
 	maps.Copy(componentSchema, accountComponent.Schema)
+
+	// account_id, type and instance make up the component's API path and resource id, so they can
+	// only change by replacement; copied because accountComponent.Schema is shared with the nested
+	// component block of sysdig_secure_cloud_auth_account, where ForceNew would replace the account
+	for _, key := range []string{SchemaType, SchemaInstance} {
+		immutable := *componentSchema[key]
+		immutable.ForceNew = true
+		componentSchema[key] = &immutable
+	}
+
 	return componentSchema
 }
 
@@ -109,7 +119,7 @@ func resourceSysdigSecureCloudauthAccountComponentUpdate(ctx context.Context, da
 	}
 
 	accountID := data.Get(SchemaAccountID).(string)
-	existingCloudAccountComponent, errStatus, err := client.GetCloudauthAccountComponentSecure(
+	_, errStatus, err := client.GetCloudauthAccountComponentSecure(
 		ctx, accountID, data.Get(SchemaType).(string), data.Get(SchemaInstance).(string))
 	if err != nil {
 		if strings.Contains(errStatus, "404") {
@@ -119,12 +129,6 @@ func resourceSysdigSecureCloudauthAccountComponentUpdate(ctx context.Context, da
 	}
 
 	newCloudAccountComponent := cloudauthAccountComponentFromResourceData(data)
-
-	// validate and reject non-updatable resource schema fields upfront
-	err = validateCloudauthAccountComponentUpdate(existingCloudAccountComponent, newCloudAccountComponent)
-	if err != nil {
-		return diag.Errorf("Error updating resource: %s", err)
-	}
 
 	_, errStatus, err = client.UpdateCloudauthAccountComponentSecure(
 		ctx, accountID, data.Get(SchemaType).(string), data.Get(SchemaInstance).(string), newCloudAccountComponent)
@@ -159,15 +163,6 @@ func resourceSysdigSecureCloudauthAccountComponentDelete(ctx context.Context, da
 /*
 This function validates and restricts any fields not allowed to be updated during resource updates.
 */
-func validateCloudauthAccountComponentUpdate(existingComponent *v2.CloudauthAccountComponentSecure, newComponent *v2.CloudauthAccountComponentSecure) error {
-	if existingComponent.Type != newComponent.Type || existingComponent.Instance != newComponent.Instance {
-		errorInvalidResourceUpdate := fmt.Sprintf("Bad Request. Updating restricted fields not allowed: %s", []string{"type", "instance"})
-		return errors.New(errorInvalidResourceUpdate)
-	}
-
-	return nil
-}
-
 func cloudauthAccountComponentFromResourceData(data *schema.ResourceData) *v2.CloudauthAccountComponentSecure {
 	cloudAccountComponent := &v2.CloudauthAccountComponentSecure{
 		AccountComponent: cloudauth.AccountComponent{
--- a/sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
@@ -48,10 +48,9 @@ func TestAccSecureCloudAuthAccountFeature(t *testing.T) {
 				ImportStateVerify: true,
 			},
 			{
-				// reordering the components set elements must not produce a diff (SSPROD-71536)
-				Config:             secureAzureWithServicePrincipalFeature(accID, tenantID, true),
-				PlanOnly:           true,
-				ExpectNonEmptyPlan: false,
+				// reordering the components set elements must not produce a diff
+				Config:   secureAzureWithServicePrincipalFeature(accID, tenantID, true),
+				PlanOnly: true,
 			},
 		},
 	})
@@ -111,7 +110,7 @@ resource "sysdig_secure_cloud_auth_account_component" "azure_service_principal_2
 				display_name              = "some-display-name-2"
 				app_display_name          = "some-app-display-name-2"
 				app_id                    = "some-app-id-2"
-				app_owner_organization_id = "some-app-owner-organization-id"
+				app_owner_organization_id = "some-app-owner-organization-id-2"
 		  }
 	  }
   })
--- a/sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
+++ b/sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
@@ -24,6 +24,9 @@ import (
 func TestAccSecureCloudAuthAccountFeature(t *testing.T) {
 	rText := func() string { return acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum) }
 	accID := rText()
+	// generated once so every step renders the same tenant id, otherwise the reorder step
+	// would also diff on provider_tenant_id
+	tenantID := acctest.RandStringFromCharSet(36, acctest.CharSetAlphaNum)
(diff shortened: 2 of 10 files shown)


sysdig/resource_sysdig_secure_cloud_auth_account_component.go
```diff
@@ -109,7 +119,7 @@ func resourceSysdigSecureCloudauthAccountComponentUpdate(ctx context.Context, da
 	}
 
 	accountID := data.Get(SchemaAccountID).(string)
-	existingCloudAccountComponent, errStatus, err := client.GetCloudauthAccountComponentSecure(
+	_, errStatus, err := client.GetCloudauthAccountComponentSecure(
```
tembleking: This pre-check GET now serves two purposes worth flagging:

1. Its only consumer (`validateCloudauthAccountComponentUpdate`) was deleted in this diff, so the result is discarded (`_`) and this is a dead extra round-trip. The identical 404 check already exists after the PUT below, so this call could be removed entirely, matching Delete's pattern (no precursor GET).
2. On 404, both here and in Read (line 101), the function returns `nil` without `data.SetId("")`. If the component was deleted out-of-band, Terraform never learns and keeps the stale resource in state forever, contrary to the convention this file follows correctly in Delete.

Suggested fix for (2), applied in both Read and here:
```go
if strings.Contains(errStatus, "404") {
    data.SetId("")
    return nil
}
```

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
```diff
@@ -46,10 +45,21 @@ func getAccountComponentSchema() map[string]*schema.Schema {
...
 			Required: true,
+			ForceNew: true,
 		},
 	}
 
 	maps.Copy(componentSchema, accountComponent.Schema)
+
+	// account_id, type and instance make up the component's API path and resource id, so they can
+	// only change by replacement; copied because accountComponent.Schema is shared with the nested
+	// component block of sysdig_secure_cloud_auth_account, where ForceNew would replace the account
```
tembleking, line 56: Good reasoning for not copying ForceNew onto the shared schema here. Two follow-ups worth considering though:

- The runtime guard that used to enforce this (`validateCloudauthAccountComponentUpdate`) was deleted in this diff, so the nested `component` block in `sysdig_secure_cloud_auth_account` (`SchemaComponent` in `resource_sysdig_secure_cloud_auth_account.go`) now has zero enforcement on `type`/`instance` changes, not ForceNew and not a manual check. No test covers what the API actually does when a nested component's identity fields change via the account resource (delete-old vs. orphan).
- `validateCloudauthAccountUpdate` in that same file still hand-rejects account-level identity field changes (`enabled`/`provider_type`/`provider_id`/`organization_id`) at Update time instead of `ForceNew`, the exact pattern this PR just replaced here. Worth applying the same fix there for consistency.

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -25,7 +26,9 @@ func TestAccSecureOrganization(t *testing.T) {
 	// Skipping the test based on this error when it occurs.
 	rText := func() string { return acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum) }
 	accID := rText()
-	organizationApiUrl := fmt.Sprintf(`%s/api/cloudauth/v1/organizations`, os.Getenv("SYSDIG_SECURE_URL"))
+	// the host is left out on purpose: SYSDIG_SECURE_URL is not set in every environment, and with
+	// it interpolated the pattern stops matching the error raised against the default endpoint
+	organizationAPIFailure := regexp.MustCompile(`POST \S*/api/cloudauth/v1/organizations giving up after \d+ attempt\(s\)`)
```
tembleking: This regex is both too loose and too strict:

- No end-anchor and permissive `\S*`/`\d+` mean it also matches retry-exhaustion on unrelated transport failures (DNS/TLS outage) against the same endpoint, reporting a real CI/infra issue as a skip instead of a failure.
- It requires `organizations` to be immediately followed by ` giving up`, which breaks when `SYSDIG_ORG_API_ASYNC=true` appends `?async=true`, or when the API returns a non-retried 4xx (go-retryablehttp only retries 5xx/429, so there's no "giving up after N attempt(s)" text at all in that case). Either way the new `return err` fallback would hard-fail instead of skip.

Suggestion: match on the specific missing-credentials error text/status rather than the generic retry-exhaustion message.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -172,39 +172,39 @@ func secureOrganizationFromResourceData(data *schema.ResourceData) *v2.Organizat
 	secureOrganization.ManagementAccountId = data.Get(SchemaManagementAccountID).(string)
 	secureOrganization.OrganizationRootId = data.Get(SchemaOrganizationRootID).(string)
 	secureOrganization.AutomaticOnboarding = data.Get(SchemaAutomaticOnboarding).(bool)
-	organizationalUnitIdsData := data.Get(SchemaOrganizationalUnitIds).([]any)
+	organizationalUnitIdsData := data.Get(SchemaOrganizationalUnitIds).(*schema.Set).List()
```
tembleking: This hand-rolls the `range set.List() { append(...string) }` loop 5 times instead of using the existing `schemaSetToList()` helper (`sysdig/tfresource.go:59`).

Suggestion: `secureOrganization.X = schemaSetToList(data.Get(SchemaX))` for each of the 5 fields, removes ~25 lines and keeps the conversion logic in one place.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -39,35 +39,35 @@ func resourceSysdigSecureOrganization() *schema.Resource {
 				Required: true,
 			},
 			SchemaOrganizationalUnitIds: {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
```
tembleking, line 50: Minor/theoretical: these TypeSet fields use the SDK's default hash with no custom `Set` function, so two distinct IDs whose hash collides would silently collapse into one element. Not exploitable in practice for real cloud IDs, just flagging the property since it has no mitigation.

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -55,13 +57,41 @@ func TestAccSecureOrganization(t *testing.T) {
...
+		[]string{"folders/444444444444", "folders/333333333333"},
+		[]string{"sample-project-two", "sample-project-one"},
+		[]string{"sample-project-four", "sample-project-three"},
+	)
+}
+
+func secureOrgConfig(accountID string, includedGroups, excludedGroups, includedAccounts, excludedAccounts []string) string {
 	// this is a base64 encoded service account key
 	test_service_account_key_encoded := getEncodedGCPServiceAccountKeyForOrg("sample", accountID)
+	quote := func(values []string) string { return `"` + strings.Join(values, `", "`) + `"` }
```
tembleking: `quote()` renders an empty/nil slice as `[""]` (one empty-string element) instead of `[]`, because `strings.Join` on empty input returns `""`. Not hit by current call sites (always 2 elements), but will bite the moment this helper is reused for an empty-list case.

Also duplicates an equivalent `%q`-join helper already in `resource_sysdig_secure_zone_rules_ordering_plan_test.go`'s `zoneExpressionConfig`. Worth factoring one shared helper instead of two slightly different ones.

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -55,13 +57,41 @@ func TestAccSecureOrganization(t *testing.T) {
...
 func secureOrgWithAccountID(accountID string) string {
+	return secureOrgConfig(accountID,
+		[]string{"folders/111111111111", "folders/222222222222"},
+		[]string{"folders/333333333333", "folders/444444444444"},
+		[]string{"sample-project-one", "sample-project-two"},
+		[]string{"sample-project-three", "sample-project-four"},
+	)
+}
+
+func secureOrgWithAccountIDReordered(accountID string) string {
```
tembleking: `secureOrgWithAccountIDReordered` hand-retypes the same 8 literals from `secureOrgWithAccountID` in reverse instead of deriving them from the original slices. If the source IDs change without mirroring here, the test silently stops exercising a genuine reorder.

Suggestion: build this from the same slices (e.g. reverse them programmatically) so drift is structurally impossible.

sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
```diff
@@ -88,15 +99,36 @@ resource "sysdig_secure_cloud_auth_account_component" "azure_service_principal"
...
+  instance                   = "secure-posture-2"
+  service_principal_metadata = jsonencode({
+	  azure = {
+		  active_directory_service_principal = {
+				id                        = "some-id-2"
+				account_enabled           = true
+				display_name              = "some-display-name-2"
+				app_display_name          = "some-app-display-name-2"
+				app_id                    = "some-app-id-2"
+				app_owner_organization_id = "some-app-owner-organization-id"
```
tembleking: Copy-paste inconsistency: this block suffixes `id`/`display_name`/`app_display_name`/`app_id` with `-2` but leaves `app_owner_organization_id` unsuffixed. Cosmetic (doesn't affect the test), but worth fixing for consistency.

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -55,13 +57,41 @@ func TestAccSecureOrganization(t *testing.T) {
 				ImportStateVerify:       true,
 				ImportStateVerifyIgnore: []string{"component"},
 			},
+			{
+				// reordering the include/exclude set elements must not produce a diff (SSPROD-71536)
+				Config:             secureOrgWithAccountIDReordered(accID),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: false,
```
tembleking: `ExpectNonEmptyPlan: false` is redundant, that's the field's zero value. Can be dropped (same redundancy exists at `resource_sysdig_secure_cloud_auth_account_feature_test.go:54`).

tembleking (changes requested):
Automated review (/code-review max). One real correctness bug (404 handling doesn't clear state on Read/Update); rest is consistency, cleanup and test-hygiene notes. Details inline.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Agree, but it is in theory. It will never happened.

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
```diff
@@ -46,10 +45,21 @@ func getAccountComponentSchema() map[string]*schema.Schema {
 		SchemaAccountID: {
 			Type:     schema.TypeString,
 			Required: true,
+			ForceNew: true,
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in a71f5a9.

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
ivanbesinovic-sysdig: Fixed in 73f7318 (both points).

sysdig/resource_sysdig_secure_organization_test.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_organization_test.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_organization_test.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_cloud_auth_account_feature_test.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_organization_test.go
ivanbesinovic-sysdig: Fixed in 73f7318.

sysdig/resource_sysdig_secure_cloud_auth_account_component.go
ivanbesinovic-sysdig: buut the nested block never had that guard - validateCloudauthAccountComponentUpdate had one call
site, the standalone Update - so it's a pre-existing gap. Agreed for provider_id/provider_type,
but `enabled` can't be ForceNew (would destroy the account to toggle a flag) and organization_id
is Computed. Follow-up for both.

sysdig/resource_sysdig_secure_organization.go
tembleking, line 50: Confirmed, this is a real gap: the resource has no `SchemaVersion`/`StateUpgraders` for this `TypeList` -> `TypeSet` change on 5 attributes. This codebase already has the identical precedent for a List->Set migration in `resource_sysdig_sso_group_mapping.go` (`team_ids`, lines 26-32 and 267-307): it bumps `SchemaVersion: 1` and adds a `StateUpgraders` entry with a V0 schema retaining the old `TypeList` shape, even though the `Upgrade` function body itself is a no-op (`return rawState, nil`). The SDK needs that V0 type to decode state written by released versions before applying the new schema; without it, existing `sysdig_secure_organization` state can fail to decode or produce a spurious diff on upgrade. Worth applying the same pattern here.

sysdig/resource_sysdig_secure_organization_test.go
```diff
@@ -19,13 +18,16 @@ import (
...
-	// XXX: TF acceptance tests for secure org onboarding need an actual existing gcp project
-	// along with an actual service_principal_key to scrape all folders and projects under the org.
-	// Without it POST /organizations call will fail with 500 error.
-	// Skipping the test based on this error when it occurs.
+	// The organization API scrapes every folder and project under the organization with the service
+	// principal from the config, so it only succeeds against a real GCP organization. Gate the test
+	// on that fixture: recognising the resulting failure by its message cannot be told apart from a
+	// genuine outage against the same endpoint.
+	if os.Getenv("SYSDIG_SECURE_GCP_ORG_TESTS") == "" {
+		t.Skip("Skipping tests on sysdig_secure_organization resource because SYSDIG_SECURE_GCP_ORG_TESTS is not set")
```
tembleking: Confirmed, this doesn't fix the underlying problem, it just moves it: `getEncodedGCPServiceAccountKeyForOrg` still returns a fully synthetic service account (`private_key_id: "xxxx..."`, a placeholder PEM key), and `secureOrgConfig` still hardcodes `organization_root_id = "organizations/123456789012"` and fake project IDs (`sample-project-one`, etc). So setting `SYSDIG_SECURE_GCP_ORG_TESTS` doesn't supply a real fixture, it just removes the skip, meaning the POST fails immediately and the reorder assertion this test exists for still never runs, with or without the flag. Needs real fixture values wired in (e.g. from env vars) before this gate does anything useful.

tembleking (changes requested):
Follow-up after the previous review's fixes (thanks for the quick turnaround on those). Verified both of Copilot's remaining comments directly against the code, both are real and worth blocking on.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 18acc41.

sysdig/resource_sysdig_secure_organization_test.go
ivanbesinovic-sysdig: Fixed in 18acc41.

sysdig/resource_sysdig_secure_organization.go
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in 18acc41.

tembleking (approved):
LGTM. All prior review findings (mine + Copilot's) are fixed or reasonably deferred, verified the state upgrader and the new GCP fixture gating directly against the code, and CI is green across the board.

PR #754, opened 2026-09-16
feat(provider): expose org API async as a provider attribute

ivanbesinovic-sysdig:
Adds `sysdig_secure_org_api_async` to the provider block so the async organization API can be enabled declaratively; `SYSDIG_ORG_API_ASYNC` keeps working, and `SYSDIG_SECURE_ORG_API_ASYNC` is accepted as the name derived from the attribute. Async is scoped to create, update and delete — reads never carry the parameter, since the read path accepts only 200. When the option is enabled, destroy waits for the organization to actually disappear instead of returning on the acknowledgement, so a replacement cannot race a deletion that is still running; a synchronous destroy is unchanged, since it is already complete when it returns.

One behaviour change applies to every configuration: the delete timeout is 30 minutes rather than 5, because it cannot be conditioned on provider configuration (`Timeouts` is static schema). For anyone already exporting `SYSDIG_ORG_API_ASYNC`, reads also drop the query parameter; that name keeps its exact-string rule, so only `SYSDIG_SECURE_ORG_API_ASYNC` accepts the conventional truthy values (`1`, `t`, `TRUE`). An update that finds the organization gone is now reported rather than recorded as a successful update.


sysdig/provider_org_api_async_test.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+		})
+	}
+}
+
+// Only the exact string "true" has ever enabled this. Anything else must stay disabled rather
+// than flip the flag or fail provider configuration on an unparseable value.
+func TestProviderOrgAPIAsyncEnvValues(t *testing.T) {
+	enabled := []string{"true"}
+	disabled := []string{"", "false", "0", "1", "t", "TRUE", "True", "yes", "on", "enabled", "true "}
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in 8d25ec8. Every subtest now calls `clearOrgAPIAsyncEnv`, which clears each name in `orgAPIAsyncEnvVars` — the same list the default func reads, so the two can't drift apart if another name is added.

sysdig/provider_org_api_async_wiring_test.go
```diff
@@ -0,0 +1,131 @@
...
+				http.MethodPost:   "async=true",
+				http.MethodPut:    "async=true",
+				http.MethodDelete: "async=true",
+				http.MethodGet:    "",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in 8d25ec8. This file was merged into `provider_org_api_async_test.go`, where every test clears both variables via `clearOrgAPIAsyncEnv`; the `nil` case was dropped as redundant with the explicit `false`. The unit suite now passes with both variables exported.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -111,6 +113,10 @@ func resourceSysdigSecureOrganizationCreate(ctx context.Context, data *schema.Re
 	if err != nil {
 		return diag.Errorf("Error creating resource: %s %s", errStatus, err)
 	}
+	// an empty id would drop the organization out of state while it exists server side
+	if orgCreated.GetId() == "" {
+		return diag.Errorf("Error creating resource: the organization was accepted but no id was returned")
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Our backend never produces that case — the async branch returns the persisted organization and the handler writes the 202 with that body, so the id is always present, and the outcome is unchanged from before this PR where a bodyless 202 failed with a parse error and left the same orphan; recovering an id without one would need a lookup by management account that the client doesn't expose yet, which is tracked with the separate create-never-calls-Read follow-up.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +139,28 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
 }
 
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, data *schema.ResourceData) error {
+	return retry.RetryContext(ctx, data.Timeout(schema.TimeoutDelete), func() *retry.RetryError {
+		_, errStatus, err := client.GetOrganizationSecure(ctx, data.Id())
+		if err != nil && strings.Contains(errStatus, "404") {
+			return nil
+		}
+		if err != nil {
+			return retry.RetryableError(fmt.Errorf("waiting for organization %s to be deleted: %s %w", data.Id(), errStatus, err))
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in e664f28 — the wait now classifies the confirmation read: 404 is success, 4xx other than 429 is `retry.NonRetryableError`, and everything else stays retryable. Note these endpoints return the status only as a status line with a plain error (`ErrorAndStatusFromResponse`), not an `APIError`, so the code is parsed off that line. `TestOrganizationDeleteWaitClassifiesErrors` asserts 401/403 report without a second read, and reverting the classification fails it.

sysdig/provider_org_api_async_test.go
```diff
@@ -0,0 +1,366 @@
...
+		t.Errorf("polled %d times, expected the delete to keep polling until the organization was gone", gets)
+	}
+}
+
+// A 202 without a usable body must not surface as a proto parse error, but create still needs an id.
+func TestOrganizationCreateRejectsAckWithoutID(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusAccepted)
+	}))
+	defer srv.Close()
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Correct, and it was my own regression from the previous commit: once create went back to a strict decode, the bodyless fixture failed at `proto: syntax error` and never reached the guard — I confirmed the diagnostic was `Error creating resource:  proto: syntax error (line 1:1)`. Fixed in 3ec0510: the fixture now returns `{}` and the test asserts the diagnostic contains "no id was returned", so removing the guard fails it.

mbarbieri (dismissed):

ivanbesinovic-sysdig:
On the three points from the latest review:

**Unconditional delete wait and 30-minute timeout** — correct, and the PR description said the opposite; it has been fixed to state that both apply to every configuration, since `Timeouts` is static schema and cannot be conditioned on provider configuration. Moving acknowledgement handling into the client, which is the only layer that knows async was requested, is tracked as follow-up.

**404 on the update itself** — fixed in 13bb9d7. It now reports `organization %s no longer exists` rather than returning nil with the id kept. Note the other half of the suggestion would not work here: clearing the id in `Update` hands Terraform an empty state for an update it planned, which it rejects with `Provider produced inconsistent result after apply … Root object was present, but now absent` — that is exactly what 89fb18e fixed on the read-back path, so both 404 paths now report instead of clearing.

**Immediate read-back observing the pre-update organization** — this one does not apply to this backend. The async branch persists the organization row synchronously before the acknowledgement is returned (`UpsertCloudOrganization`, then the NATS publish, then `GetCloudOrganization` for the response body); only the member-account scrape is deferred, and member accounts are not part of this resource's schema. So an immediate read returns the updated row, not a stale one.

sysdig/provider_org_api_async_test.go
```diff
@@ -0,0 +1,426 @@
...
+	config := map[string]any{
+		"sysdig_secure_url":       srv.URL,
+		"sysdig_secure_api_token": "fake-token",
+		orgAPIAsyncKey:            true,
+	}
+	clients := &sysdigClients{ctx: context.Background(), d: providerData(t, config)}
+	data := schema.TestResourceDataRaw(t, resourceSysdigSecureOrganization().Schema, map[string]any{})
+	data.SetId(orgID)
+
+	if diags := resourceSysdigSecureOrganizationDelete(context.Background(), data, clients); diags.HasError() {
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: The timeout is not zero: `schema.TestResourceDataRaw` leaves `d.timeouts` nil, and `ResourceData.Timeout` returns a 20-minute `defaultTimeout` in that case (`helper/schema/resource_data.go:556-559`). I measured it — 20m0s — so the wait cannot exhaust before the third read, which is why the test is stable at ~1.5s.

The split you suggest is already in place: `TestOrganizationDeleteWaitTimesOut` calls the wait with an explicit one-second timeout, and `TestOrganizationDeleteTimeoutBudget` separately asserts the resource declares 30 minutes for delete and 5 for create. You're right that this test runs on the SDK fallback rather than the production value; that is deliberate, since propagation is covered by the second test.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,16 +215,25 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
...
+		// planned, which it rejects as an inconsistent result
 		if strings.Contains(errStatus, "404") {
-			return nil
+			return diag.Errorf("Error updating resource: organization %s no longer exists", orgID)
 		}
 		return diag.Errorf("Error updating resource: %s %s", errStatus, err)
 	}
 
-	return nil
+	diags := resourceSysdigSecureOrganizationRead(ctx, data, i)
```
tembleking: There's no wait-for-completion loop for Update, unlike Delete's `waitForOrganizationDeletion`. On the async path the 202 comes back before the change is actually applied, so the read-back right after can hand Terraform stale pre-update data. That either triggers an inconsistent-apply error or silently writes stale values into state. Given Delete already solves this, Update probably needs the same poll-until-applied treatment.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +141,50 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+		if err == nil {
+			return retry.RetryableError(fmt.Errorf("organization %s has not been deleted yet; the deletion may also have failed server side", orgID))
+		}
+
+		code := statusCodeFromStatus(errStatus)
+		if code == http.StatusNotFound || strings.Contains(errStatus, "404") {
+			return nil
+		}
+		// waiting out the delete timeout will not fix a rejected request
+		if code >= 400 && code < 500 && code != http.StatusTooManyRequests {
```
tembleking: This wait loop treats every 4xx except 429 as permanent, but the transport's `CheckRetry` already retries 409 (`client.go:293-306`). If the confirmation GET hits 409 after the transport's own retries are exhausted, `destroy` aborts right away, which is exactly the case this loop was built to handle. 409 should probably be retryable here, and 410 treated as success the same way 404 is.

sysdig/internal/client/v2/organization.go
```diff
@@ -42,8 +43,7 @@ func (c *Client) CreateOrganizationSecure(ctx context.Context, org *Organization
 	}
 
 	organization = &OrganizationSecure{}
-	err = c.unmarshalCloudauthProto(response.Body, organization)
-	if err != nil {
+	if err = c.unmarshalCloudauthProto(response.Body, organization); err != nil {
```
tembleking: `UpdateOrganizationSecure` now tolerates an empty-body async ack, but `CreateOrganizationSecure` still decodes strictly through `unmarshalCloudauthProto`. If Create's async ack can also come back empty, you get a raw decode error instead of the "accepted but no id" guard the resource already handles. Might be worth giving Create the same tolerant path.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +141,50 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+// the timeout is passed in rather than read from the resource so it can be exercised in tests;
+// the SDK has already put the same deadline on ctx
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, orgID string, timeout time.Duration) error {
+	return retry.RetryContext(ctx, timeout, func() *retry.RetryError {
+		_, errStatus, err := client.GetOrganizationSecure(ctx, orgID)
+		if err == nil {
+			return retry.RetryableError(fmt.Errorf("organization %s has not been deleted yet; the deletion may also have failed server side", orgID))
+		}
+
+		code := statusCodeFromStatus(errStatus)
```
tembleking: `statusCodeFromStatus` returns 0 whenever `errStatus` is empty, whether that's a network blip or a bad body. Zero falls straight into the retryable branch, so a genuinely broken client just polls for the full 30 minutes before reporting a "may have failed server side" message that doesn't match what actually happened. A parse failure should probably fail fast instead.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +141,50 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
 		return diag.Errorf("Error deleting resource: %s %s", errStatus, err)
 	}
 
+	// the request may only have been acknowledged, and a replacement cannot be created until the
+	// organization is actually gone
+	if err := waitForOrganizationDeletion(ctx, client, data.Id(), data.Timeout(schema.TimeoutDelete)); err != nil {
```
tembleking: Heads up on scope here: the delete-confirmation wait and the 5m to 30m timeout bump apply to every configuration, not gated by `sysdig_secure_org_api_async`, since `Timeouts` is static schema. So even someone who never touches the async flag now gets an extra GET after every delete, and a real failure takes up to 30 minutes to surface instead of 5. Worth calling out in the changelog at minimum, or gating behind the flag if that's feasible.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,16 +215,25 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
 	}
 
 	org := secureOrganizationFromResourceData(data)
+	orgID := data.Id()
 
-	_, errStatus, err := client.UpdateOrganizationSecure(ctx, data.Id(), org)
+	_, errStatus, err := client.UpdateOrganizationSecure(ctx, orgID, org)
```
tembleking: Similar scope issue on Update: it now throws away the object the PUT already returned and does a second GET no matter what, doubling the network calls even on the plain synchronous path, still inside the same 5-minute timeout. A slow PUT could eat enough of that budget that the trailing GET times out, reporting a read failure for an update that actually went through fine.

sysdig/provider.go
```diff
@@ -272,3 +279,18 @@ func (p *SysdigProvider) providerConfigure(ctx context.Context, d *schema.Resour
...
+// checked in order; the second name shipped before the attribute existed and has to keep working
+var orgAPIAsyncEnvVars = []string{"SYSDIG_SECURE_ORG_API_ASYNC", "SYSDIG_ORG_API_ASYNC"}
+
+// not schema.MultiEnvDefaultFunc: that returns the raw string, and a value the SDK cannot coerce
+// fails provider configuration for everyone already exporting one of these
+func orgAPIAsyncDefaultFunc() (any, error) {
+	for _, name := range orgAPIAsyncEnvVars {
+		if v := os.Getenv(name); v != "" {
+			enabled, err := strconv.ParseBool(v)
+			return err == nil && enabled, nil
```
tembleking: `orgAPIAsyncDefaultFunc` stops at the first non-empty env var even when `ParseBool` fails on it, so it never reaches the legacy variable, contradicting the comment that says the legacy one has to keep working. Someone with a typo'd `SYSDIG_SECURE_ORG_API_ASYNC="tru"` and a correctly-set legacy `SYSDIG_ORG_API_ASYNC=true` would silently end up with async disabled. It should only short-circuit on a successful parse.

sysdig/provider.go
```diff
@@ -272,3 +279,18 @@ func (p *SysdigProvider) providerConfigure(ctx context.Context, d *schema.Resour
...
+
+// checked in order; the second name shipped before the attribute existed and has to keep working
+var orgAPIAsyncEnvVars = []string{"SYSDIG_SECURE_ORG_API_ASYNC", "SYSDIG_ORG_API_ASYNC"}
+
+// not schema.MultiEnvDefaultFunc: that returns the raw string, and a value the SDK cannot coerce
+// fails provider configuration for everyone already exporting one of these
+func orgAPIAsyncDefaultFunc() (any, error) {
+	for _, name := range orgAPIAsyncEnvVars {
+		if v := os.Getenv(name); v != "" {
+			enabled, err := strconv.ParseBool(v)
```
tembleking: This one's a real backcompat break. Moving from `== "true"` to `ParseBool` means any legacy `SYSDIG_ORG_API_ASYNC` value that was truthy-but-not-exactly-"true", like `"1"` or `"TRUE"`, used to be a no-op and now silently turns on real async Create/Update/Delete behavior after upgrade, with zero config change on the user's end. `website/docs/index.md` doesn't mention this. Either keep the strict "true" check, or call this out loudly as a breaking change in the release notes.

tembleking:
On `sysdig/internal/client/v2/organization.go:34` (pre-existing code, not touched by this diff, so couldn't anchor inline, and not a blocker for this PR): the deferred `Body.Close()` error handler overwrites the named `err` return. In `CreateOrganizationSecure`, if decoding succeeds and the org has a real id, but `Close()` then fails, `err` becomes the close error, the function returns early, and `SetId` never runs. The org exists on the server but Terraform has no record of it, so the next apply creates a second one. The same pattern also affects Update/Delete's error reporting. Only assigning the close error when `err` is still nil would fix this. Flagging it here since this PR's async handling makes the Create case more likely to bite (slower responses, more retry paths touching the body), but it's a pre-existing issue, not something this PR needs to fix.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: No wait needed — the PUT returns the persisted organization. Fixed in 3f4cc74.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/provider.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

sysdig/provider.go
ivanbesinovic-sysdig: Fixed in 3f4cc74.

ivanbesinovic-sysdig:
Fixed in 3f4cc74.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,15 +232,24 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
...
+			return diag.Errorf("Error updating resource: organization %s no longer exists", orgID)
 		}
 		return diag.Errorf("Error updating resource: %s %s", errStatus, err)
 	}
 
+	// the response already carries the persisted organization, so state comes from it rather than
+	// from a second read that would cost another call and could outlive the update timeout
+	if err := secureOrganizationToResourceData(data, updated); err != nil {
+		return diag.FromErr(err)
+	}
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Done in cb782ad.

sysdig/provider.go
```diff
@@ -272,3 +279,32 @@ func (p *SysdigProvider) providerConfigure(ctx context.Context, d *schema.Resour
...
+// coerce fails provider configuration for everyone already exporting one of these
+func orgAPIAsyncDefaultFunc() (any, error) {
+	for _, env := range orgAPIAsyncEnvVars {
+		v := os.Getenv(env.name)
+		if v == "" {
+			continue
+		}
+		if env.strict {
+			return v == "true", nil
+		}
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Not valid — the legacy name must stay strict, or a provider upgrade flips anyone exporting `SYSDIG_ORG_API_ASYNC=1` from sync to async. The description was stale and is corrected.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +143,65 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
 		return diag.Errorf("Error deleting resource: %s %s", errStatus, err)
 	}
 
+	// only an async delete is merely acknowledged; a synchronous one is done when it returns
+	if i.(SysdigClients).orgAPIAsyncEnabled() {
+		if err := waitForOrganizationDeletion(ctx, client, data.Id(), data.Timeout(schema.TimeoutDelete)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Not valid — the gating is intentional: a synchronous delete is already complete when it returns. The description was stale and is corrected.

mbarbieri (dismissed):

draraksysdig (dismissed):

sysdig/internal/client/v2/organization.go
```diff
@@ -26,12 +27,13 @@ func (c *Client) CreateOrganizationSecure(ctx context.Context, org *Organization
...
 
-	response, err := c.requester.Request(ctx, http.MethodPost, c.organizationsURL(), payload)
+	response, err := c.requester.Request(ctx, http.MethodPost, c.withAsync(c.organizationsURL()), payload)
 	if err != nil {
 		return nil, "", err
 	}
 	defer func() {
-		if dErr := response.Body.Close(); dErr != nil {
+		// a close failure here would discard a successful create and leave the organization untracked
+		if dErr := response.Body.Close(); dErr != nil && err == nil {
```
tembleking: The new `&& err == nil` guard stops a close error from clobbering a real earlier error, but it doesn't stop the other direction: if decoding already succeeded here (the create went through, org has a real id) and `Close()` then fails, `err` still flips from nil to the close error. `CreateOrganizationSecure` returns that error, the resource never calls `SetId`, and the organization exists on the server with Terraform never having recorded it. The new comment above this line names the exact failure mode without preventing it. Logging the close failure instead of returning it, or returning the organization alongside a warning, would keep a real create from vanishing.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,15 +232,27 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
...
-			return nil
+			return diag.Errorf("Error updating resource: organization %s no longer exists", orgID)
 		}
 		return diag.Errorf("Error updating resource: %s %s", errStatus, err)
 	}
 
+	// the response already carries the persisted organization, so state comes from it rather than
+	// from a second read that would cost another call and could outlive the update timeout;
+	// an acknowledgement without one would wipe the planned values, which the plan already holds
+	if updated.GetId() != "" {
```
tembleking: Checking `updated.GetId() != ""` assumes any response with an id is a full organization, but a partial ack (just an id, everything else zero) would pass this guard just as easily as a complete body. That runs `secureOrganizationToResourceData` over it and wipes real values like `management_account_id` back to empty, with nothing surfacing an error. Worth checking whether the async PUT can ever ack with a partial body the same way the empty-body case already got a guard for.

sysdig/internal/client/v2/organization.go
```diff
@@ -74,17 +76,18 @@ func (c *Client) GetOrganizationSecure(ctx context.Context, orgID string) (organ
...
 	defer func() {
-		if dErr := response.Body.Close(); dErr != nil {
+		// a close failure must not overwrite a real error, nor fail an otherwise successful call
+		if dErr := response.Body.Close(); dErr != nil && err == nil {
 			err = fmt.Errorf("unable to close response body: %w", dErr)
 		}
 	}()
 
-	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
+	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK && response.StatusCode != http.StatusAccepted {
```
tembleking: This accepts 202 regardless of whether the client actually asked for async behavior. If the server ever answers a delete with 202 when `?async=true` wasn't on the URL, `orgAPIAsyncEnabled()` back in the resource still reads false, `waitForOrganizationDeletion` gets skipped, and `destroy` returns while the deletion may still be running server-side. Tying the 202 acceptance to the same async flag that gates the wait would close that gap.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -143,6 +211,7 @@ func resourceSysdigSecureOrganizationRead(ctx context.Context, data *schema.Reso
 	org, errStatus, err := client.GetOrganizationSecure(ctx, data.Id())
 	if err != nil {
 		if strings.Contains(errStatus, "404") {
```
tembleking: `classifyDeletionProbe` treats 410 the same as 404 for the delete-confirmation poll, but this Read check and the equivalent one in Delete (line 140) only look for `"404"`. A server that answers a gone organization with 410 instead of 404 would leave Read holding a stale id in state and make Delete error out on something that's already gone, even though the new wait logic already knows how to handle 410 correctly.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +143,65 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+	case code == http.StatusNotFound || code == http.StatusGone:
+		return nil
+	// 409 is what the transport itself retries, so it is transient here too
+	case code == http.StatusConflict || code == http.StatusTooManyRequests || code >= 500:
+		return retry.RetryableError(fmt.Errorf("confirming deletion of organization %s: %s %w", orgID, errStatus, err))
+	case code >= 400:
+		return retry.NonRetryableError(fmt.Errorf("confirming deletion of organization %s: %s %w", orgID, errStatus, err))
+	}
+
+	// no status at all: a transport failure can pass, anything else (an unreadable body, say)
```
tembleking: This fallback assumes reaching here means there was no HTTP status at all, but it also catches a decode failure on a 200 response, or an unexpected 2xx/3xx status, neither of which is really the same situation as a transport-level failure. Both get classified as non-retryable and abort the 30-minute wait immediately, when a body-read blip on an otherwise healthy 200 is exactly the kind of thing a retry a moment later would likely clear.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -131,9 +143,65 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+}
+
+// nil means the organization is gone; anything retryable is worth another poll, and everything
+// else has to be reported instead of waiting out the whole delete timeout.
+func classifyDeletionProbe(orgID, errStatus string, err error) *retry.RetryError {
+	switch code := statusCodeFromStatus(errStatus); {
+	case code == http.StatusNotFound || code == http.StatusGone:
+		return nil
+	// 409 is what the transport itself retries, so it is transient here too
+	case code == http.StatusConflict || code == http.StatusTooManyRequests || code >= 500:
```
tembleking: Bucketing every `>= 500` as retryable disagrees with the transport's own policy, which excludes 501 from its retryable set. A 501 here gets polled for the full 30 minutes instead of failing fast. The `url.Error` fallback further down has the same issue: some of what it wraps, like a malformed URL, is deterministic and will never resolve no matter how many times it's retried.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -111,6 +119,10 @@ func resourceSysdigSecureOrganizationCreate(ctx context.Context, data *schema.Re
 	if err != nil {
 		return diag.Errorf("Error creating resource: %s %s", errStatus, err)
 	}
+	// an empty id would drop the organization out of state while it exists server side
+	if orgCreated.GetId() == "" {
```
tembleking, line 124: Create fails outright when the response carries no id, while Update and Delete were both changed to tolerate an async ack with no body at all. If the real Create endpoint acks the same bodyless way under `sysdig_secure_org_api_async=true`, every create on a large organization, the exact case this flag exists for, would fail on every apply even though the organization is actually being created server-side. Might be worth confirming with the API team whether Create's async ack can also come back empty.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -2,17 +2,25 @@ package sysdig
...
 	cloudauth "github.com/draios/terraform-provider-sysdig/sysdig/internal/client/v2/cloudauth/go"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourceSysdigSecureOrganization() *schema.Resource {
 	timeout := 5 * time.Minute
+	// an async delete waits out the member-account cascade, which outlasts the other operations
+	deleteTimeout := 30 * time.Minute
```
tembleking, line 24: This raises the Delete timeout from 5 to 30 minutes for every configuration, not just when async is enabled, since `Timeouts` is fixed at schema-definition time and can't branch on a resource attribute. A synchronous delete that genuinely hangs now takes 6x longer to be reported as failed. Also worth a line in the docs, since right now only the async section mentions 30 minutes.

sysdig/provider_org_api_async_test.go
```diff
@@ -0,0 +1,530 @@
...
+					t.Errorf("%s query = %q, want %q", method, got, want)
+				}
+			}
+		})
+	}
+}
+
+// An async delete is only acknowledged, so destroy has to wait for the organization to go away.
+// Driven through the resource so that dropping the wait from Delete fails here.
+func TestOrganizationDeleteWaitsForRemoval(t *testing.T) {
```
tembleking, line 235: `resourceSysdigSecureOrganizationDelete` is exercised here through `schema.TestResourceDataRaw`, which only takes the schema map and never attaches `.Timeouts`. That means `data.Timeout(schema.TimeoutDelete)` in this test falls back to the SDK's generic 20-minute default rather than the resource's actual 30-minute value. If the real timeout regressed back to 5 minutes, or got wired to the wrong key, this test wouldn't catch it. Only `TestOrganizationDeleteTimeoutBudget` checks the value itself, and it does so statically without exercising the wait.

website/docs/index.md
```diff
@@ -151,6 +151,18 @@ When Secure resources are to be created, this authentication must be in place.
   on-prem installations. It can also be sourced from the `SYSDIG_SECURE_INSECURE_TLS`
   environment variable. By default, this is false.<br/><br/>
 
+* `sysdig_secure_org_api_async` - (Optional) Creates and updates
```
tembleking, line 154: This documents `SYSDIG_SECURE_ORG_API_ASYNC` but never mentions that the legacy `SYSDIG_ORG_API_ASYNC` still works too, with a stricter exact-match rule and a narrower scope than before. Someone who already has the legacy variable set has no way to learn any of that from this section.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/provider_org_api_async_test.go
ivanbesinovic-sysdig: Fixed in a593160.

website/docs/index.md
ivanbesinovic-sysdig: Fixed in a593160.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Confirmed on the API side: the async path persists the organization and re-reads it before answering, and the handler writes the 202 together with that body — so the create ack is never bodyless or partial. The guard stays as a safety net. Same check covers the partial-body case above, where the guard now also requires the management account.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Docs added in a593160 (resource page now has a Timeouts section). The 30m stays: `Timeouts` is static schema, and a synchronous delete is cut by the gateway at ~100s long before either budget applies.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,96 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+
+	// without a status the failure is transport level or a body that could not be read; only the
+	// deterministic ones are hopeless, and the transport already knows which those are
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) {
+		if transient, _ := retryablehttp.DefaultRetryPolicy(ctx, nil, urlErr); !transient {
+			return fatal()
+		}
+	}
+	return retryable()
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in b903d65 — along with the bodyless-202 finding from the same review; both are gated the same way now.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,15 +263,27 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
...
+		if organizationIsGone(errStatus) {
+			return diag.Errorf("Error updating resource: organization %s no longer exists", orgID)
 		}
 		return diag.Errorf("Error updating resource: %s %s", errStatus, err)
 	}
 
+	// the response already carries the persisted organization, so state comes from it rather than
+	// from a second read that would cost another call and could outlive the update timeout;
+	// a bare acknowledgement lacks the required management account and would wipe planned values
+	if updated.GetId() != "" && updated.GetManagementAccountId() != "" {
```
tembleking, line 281: The extra `managementAccountId` check narrows this a bit, but a bare ack that happens to include both `id` and `managementAccountId` while leaving everything else empty would still sail through. `secureOrganizationToResourceData` runs unconditionally at that point, so `organizational_unit_ids`, the included/excluded cloud accounts, and any org groups can still get wiped from state with nothing surfacing an error. Worth checking whether the real async PUT can ack with just those two fields populated.

sysdig/internal/client/v2/organization.go
```diff
@@ -97,41 +94,69 @@ func (c *Client) UpdateOrganizationSecure(ctx context.Context, orgID string, org
...
+}
+
+func (e *MalformedBodyError) Unwrap() error { return e.Err }
+
+func (c *Client) unmarshalOrganizationBody(response *http.Response, organization *OrganizationSecure) error {
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
+	if response.StatusCode == http.StatusAccepted && len(bytes.TrimSpace(body)) == 0 {
```
tembleking, line 134: This only tolerates an empty body when the status is exactly 202, but Create and Update's own success check a few lines up also accepts 200 and 201. If the real API ever acknowledges an async create or update with 200 or 201 and an empty body instead of 202, this treats it as a malformed payload and fails the call even though the server accepted it.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,101 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+	// without a status the failure is transport level or a body problem; only the deterministic
+	// ones are hopeless: a body that does not parse will not parse on the next attempt either,
+	// and for the transport the same classification the request layer uses already knows which
+	var malformed *v2.MalformedBodyError
+	if errors.As(err, &malformed) {
+		return fatal()
+	}
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) {
+		if transient, _ := retryablehttp.DefaultRetryPolicy(ctx, nil, urlErr); !transient {
```
tembleking, line 215: The outer `ctx` already carries the same 30-minute deadline this wait uses internally, since the SDK sets it before this function even starts. If a transient network error lands right around that boundary, `ctx.Err()` is already non-nil, so `DefaultRetryPolicy` reports it as non-transient and this gets classified as fatal instead of a plain timeout. The operation fails either way, but the error message ends up blaming a connection reset instead of saying the wait just ran out.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,101 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+			return diag.FromErr(err)
+		}
+	}
+
 	return nil
 }
 
+// the timeout is passed in rather than read from the resource so it can be exercised in tests;
+// the SDK has already put the same deadline on ctx
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, orgID string, timeout time.Duration) error {
```
tembleking, line 159: Each `GetOrganizationSecure` call inside this loop already goes through the transport's own retry policy (up to 5 attempts, backing off as long as 30 seconds each) before `classifyDeletionProbe` even gets a chance to decide the next tick. A stretch of transient 503s or 409s during a real delete cascade can burn a good chunk of the 30-minute budget on redundant backoff rather than actual progress checks, making the wait more likely to time out on an organization that was deleting just fine.

sysdig/internal/client/v2/organization.go
```diff
@@ -97,41 +94,69 @@ func (c *Client) UpdateOrganizationSecure(ctx context.Context, orgID string, org
...
 
 	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusCreated && response.StatusCode != http.StatusAccepted {
 		errStatus, err := c.ErrorAndStatusFromResponse(response)
 		return nil, errStatus, err
 	}
 
 	organization = &OrganizationSecure{}
-	err = c.unmarshalCloudauthProto(response.Body, organization)
-	if err != nil {
+	if err := c.unmarshalOrganizationBody(response, organization); err != nil {
```
tembleking, line 111: This uses `:=` for the decode error, unlike the same line in Create and Get, which assign to the outer named-return `err` with `=`. Nothing depends on that today since every branch here returns explicit values, but it's the exact named-return-plus-defer shape that caused the close-error bug fixed elsewhere in this file. A later change that adds a naked return or another deferred check to Update would silently miss a decode failure because of this shadow.

tembleking:
None of the new async/wait/ack behavior in this PR is exercised by the acceptance test suite for this resource; `sysdig/resource_sysdig_secure_organization_test.go` (tagged `tf_acc_sysdig_secure`) has no changes in this diff. Everything new is validated only against `httptest` fakes built on assumptions about how the real cloudauth API responds to async requests. If those assumptions are off, for example the delete cascade always answering 202 regardless of whether `?async=true` was sent, it would only surface against a real large-organization apply/destroy, not in CI.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in ce69aea — no field checks left: an acknowledgement now reports no organization at all, so nothing it carries can reach state.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in ce69aea.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in ce69aea.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in ce69aea.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in ce69aea.

ivanbesinovic-sysdig:
Fixed in ce69aea — added `TestAccSecureOrganizationAsync`, which runs create, update and destroy with the option enabled. It lives in the acceptance suite, so it needs the GCP org fixture and does not run in PR CI.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -163,15 +273,27 @@ func resourceSysdigSecureOrganizationUpdate(ctx context.Context, data *schema.Re
...
 	}
 
+	// a persisted organization comes back from the update itself, so state comes from it rather
+	// than from a second read that would cost another call and could outlive the update timeout;
+	// an acknowledgement carries no organization, and the plan already holds what was sent
+	if updated != nil {
+		if err := secureOrganizationToResourceData(data, updated); err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in ba21994 — the update no longer reads its answer into state at all, so no body can overwrite what was applied; Read is what brings the server view in.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -34,7 +43,7 @@ func resourceSysdigSecureOrganization() *schema.Resource {
 			Create: schema.DefaultTimeout(timeout),
 			Update: schema.DefaultTimeout(timeout),
 			Read:   schema.DefaultTimeout(timeout),
-			Delete: schema.DefaultTimeout(timeout),
+			Delete: schema.DefaultTimeout(deleteTimeout),
```
tembleking: Orgs that are already in state won't pick up this new delete default. SDK v2 stores timeouts in private state at create time, and a destroy reuses the prior private state. So an org created by an older release keeps `delete = 5m`, and the async wait gives up after 5 minutes. Adding `timeouts { delete = "60m" }` doesn't help either: without an attribute diff it plans as "No changes" and the stored value stays. I reproduced this with Terraform 1.16. An org created on master failed its destroy at about 5m, while a control org created with this build went through.

Existing users with `SYSDIG_ORG_API_ASYNC=true` would hit this on their first destroy after upgrading. Could the wait budget be computed by the resource itself, maybe via `DeleteWithoutTimeout`? If that's out of scope, the docs at `secure_organization.md:41` and `index.md:162` should at least mention that existing orgs need an apply with a real change first.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,111 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+const organizationDeletionProbeTimeout = 10 * time.Second
+
+// the timeout is passed in rather than read from the resource so it can be exercised in tests;
+// the SDK has already put the same deadline on ctx
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, orgID string, timeout time.Duration) error {
+	return retry.RetryContext(ctx, timeout, func() *retry.RetryError {
+		probeCtx, cancel := context.WithTimeout(ctx, organizationDeletionProbeTimeout)
+		defer cancel()
+
+		_, errStatus, err := client.GetOrganizationSecure(probeCtx, orgID)
```
tembleking: This probe runs a GET without `verbosity`. The backend reads that as DETAIL and calls `ListCloudAccounts` for every member. For a 5000-account org, each probe comes to ~10MB, ~1.7s of CPU and ~200MB allocated, and the provider also dumps and regex-redacts the whole body. A 10-minute cascade means ~60 of those just to learn the org still exists, and it hits the large orgs this option is meant for.

```
GET /cloudauth/v1/organizations/{id}?verbosity=VERBOSITY_IDENT
```
IDENT returns the same 200/404 from the org row alone.

sysdig/sysdig_clients.go
```diff
@@ -365,6 +370,12 @@ func (c *sysdigClients) commonClientV2() (v2.Common, error) {
 	return c.commonV2, err
 }
 
+// the organization resource has to know this to decide whether a delete needs confirming
+func (c *sysdigClients) orgAPIAsyncEnabled() bool {
+	enabled, _ := c.d.Get("sysdig_secure_org_api_async").(bool)
```
tembleking: This reads the shared provider `ResourceData` without holding `c.mu`. The SDK's `DiffFieldReader` memoizes reads in a plain map (`field_reader_diff.go:55/84`), so when another resource's first `GetClientType()` writes that map at the same moment, the plugin dies with `fatal error: concurrent map read and map write`. I got it 5/5 when the calls overlapped, and `-race` flags it. The flag is already in the cached client config through `WithOrgAPIAsync`, so reading it from there avoids the shared state completely. Taking `c.mu` would work too.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,111 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+		return retry.NonRetryableError(fmt.Errorf("confirming deletion of organization %s: %s %w", orgID, errStatus, err))
+	}
+
+	if organizationIsGone(errStatus) {
+		return nil
+	}
+	if code := statusCodeFromStatus(errStatus); code != 0 {
+		switch {
+		// the same set the transport itself retries: 409, 429 and 5xx other than 501
+		case code == http.StatusConflict, code == http.StatusTooManyRequests:
```
tembleking: Because the probe goes through the full client, the transport retries (1/2/4/8s backoff) run inside the 10s cap. Against a persistent 503, the ctx expires before the client returns a status, so `errStatus` is empty and the user only sees `context deadline exceeded`. As a result, the 409/429/5xx branches below never execute. In a 45s run I saw 16 GETs over 4 new connections. Also, a 200 with an HTML body (LB, WAF, maintenance page) ends up as `MalformedBodyError`, which is fatal. That fails the destroy even though a 200 proves the org still exists.

A status-only check with transport retries disabled would fix both. One option is a ctx marker that `CheckRetry` honors. That also makes `MalformedBodyError` and its branch unnecessary.

sysdig/internal/client/v2/organization.go
```diff
@@ -54,37 +55,35 @@ func (c *Client) GetOrganizationSecure(ctx context.Context, orgID string) (organ
...
-	}()
-
-	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
+	// a close failure says nothing the caller can act on, and returning it would discard a
+	// successful call: a created organization would exist server side with nothing tracking it
+	defer func() { _ = response.Body.Close() }()
+
+	// 202 is only expected when async was requested: accepting it otherwise would let a destroy
+	// return on an acknowledgement while the resource, seeing the flag off, skips the wait
+	acknowledged := c.config.secureOrgAPIAsync && response.StatusCode == http.StatusAccepted
```
tembleking: cloudauth's `DeleteOrganization` always answers `204 No Content`. In async mode it just spawns the goroutine (`http_handler.go:1454`, `service/api.go:1879-1883`), so `response.StatusCode == http.StatusAccepted` can never be true here. The three tests that answer the DELETE with 202 cover a response the backend doesn't send, and nothing covers the real async 204. Right now the wait works because it's keyed on the flag. If the follow-up that moves ack handling into the client keys on this status instead, destroy would return without polling while the cascade is still running, and all 31 org tests would stay green (I checked). Dropping the 202 branch and adding an async + 204 test would lock in the real behavior.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,111 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
 
+// a confirmation read carries the transport's own retries, which back off as far as 30 seconds
+// each; without a cap one unlucky probe spends that inside a single tick of the wait, burning the
+// delete budget on backoff rather than on checking whether the cascade finished
+const organizationDeletionProbeTimeout = 10 * time.Second
+
+// the timeout is passed in rather than read from the resource so it can be exercised in tests;
+// the SDK has already put the same deadline on ctx
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, orgID string, timeout time.Duration) error {
+	return retry.RetryContext(ctx, timeout, func() *retry.RetryError {
```
tembleking: The only exit from this loop is a 404. The async cascade runs in a goroutine whose error is logged and dropped. For example, an FK violation can happen if destroy overlaps an onboarding job that's still adding members. When the cascade rolls back, the GET keeps returning 200, and the destroy spins for the whole budget. Once the budget runs out, `RetryContext` returns the last probe error instead of the timeout. The user then sees `confirming deletion of organization X: Get "...": context deadline exceeded`, which points at the 10s probe cap and hides the real cause.

```go
if errors.Is(err, context.DeadlineExceeded) && ctx.Err() != nil {
    return fmt.Errorf("organization %s not confirmed deleted within %s", id, budget)
}
```
Re-issuing the DELETE after a while might also be worth considering.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +138,111 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+	// and for the transport the same classification the request layer uses already knows which
+	var malformed *v2.MalformedBodyError
+	if errors.As(err, &malformed) {
+		return fatal()
+	}
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) {
+		// asked with a fresh context on purpose: the wait's own deadline, once expired, makes the
+		// policy report any error as permanent and the failure would blame the network instead
+		if transient, _ := retryablehttp.DefaultRetryPolicy(context.Background(), nil, urlErr); !transient {
```
tembleking: `DefaultRetryPolicy` gets the outer `*url.Error` here. Its `Err` is retryablehttp's "giving up after N attempt(s)" wrapper, so `isCertError`'s direct type assertion never matches. A real `certificate signed by unknown authority` is classified as retryable. I confirmed that with `httptest.NewTLSServer` and the real v2 client. If a cert expires, or a TLS-inspecting proxy shows up during the wait, destroy polls for the full budget instead of failing fast. The fatal-cert test at `provider_org_api_async_test.go:511` passes only because it builds a single-level `url.Error` by hand.

```go
var certErr *tls.CertificateVerificationError
if errors.As(err, &certErr) { return fatal }
```

website/docs/index.md
```diff
@@ -151,6 +151,21 @@ When Secure resources are to be created, this authentication must be in place.
...
+  `sysdig_secure_organization` asynchronously, so the call returns as soon as the request
+  is accepted instead of waiting for every member account; member accounts then appear
```
tembleking: This section reads as a pure latency trade-off, but async also changes what failures look like. In sync mode, a failed onboarding (e.g. the GCP principal lacks folder permissions) returns 500 and the apply fails. In async mode, the POST returns 202, and the NATS job's error is only logged and acked (`api/nats_handler.go:152-159`). Terraform reports success, the plan stays clean, and the org never onboards a member account. Separately, a `PUT ?async=true` sent while an earlier onboarding job holds the org row lock blocks until that job commits, and it can run past the unchanged 5m Update timeout. Users picking async for large orgs should know both.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e — took the documentation alternative you offered: both `secure_organization.md` and `index.md` now say that an organization created by an earlier version keeps its stored 5m delete timeout until an apply that changes the resource records the new one, and that a `timeouts` block alone produces no change.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/sysdig_clients.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/internal/client/v2/organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/resource_sysdig_secure_organization.go
ivanbesinovic-sysdig: Fixed in 5a9c70e.

website/docs/index.md
ivanbesinovic-sysdig: Fixed in 5a9c70e.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +139,119 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+// spend; the next tick of the wait asks again
+const organizationDeletionProbeTimeout = 10 * time.Second
+
+// the budget is passed in rather than read from the resource so it can be exercised in tests
+func waitForOrganizationDeletion(ctx context.Context, client v2.OrganizationSecureInterface, orgID string, budget time.Duration) error {
+	var fatal bool
+
+	err := retry.RetryContext(ctx, budget, func() *retry.RetryError {
+		probeCtx, cancel := context.WithTimeout(ctx, organizationDeletionProbeTimeout)
+		defer cancel()
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in f3d6562.

sysdig/resource_sysdig_secure_organization.go
```diff
@@ -125,15 +139,125 @@ func resourceSysdigSecureOrganizationDelete(ctx context.Context, data *schema.Re
...
+
+	err := retry.RetryContext(budgetCtx, budget, func() *retry.RetryError {
+		probeCtx, cancel := context.WithTimeout(budgetCtx, organizationDeletionProbeTimeout)
+		defer cancel()
+
+		exists, errStatus, err := client.OrganizationExistsSecure(probeCtx, orgID)
+		switch {
+		case err != nil:
+			probe := classifyDeletionProbe(orgID, errStatus, err)
+			fatal = probe != nil && !probe.Retryable
```
tembleking: `fatal` is written here from the refresh goroutine and read at line 192 after `RetryContext` returns, with no synchronization. `WaitForStateContext` returns on `ctx.Done()` without waiting for a refresh that's still in flight, so a probe cancelled at the deadline (or by Ctrl-C) can still write `fatal` while the caller reads it. `go test -race` reports it on every run with a blocking fake, and `just testacc` and the CI matrix both run with `-race`. Without `-race`, a 403 that lands at the deadline came back in three different shapes (unwrapped, wrapped, or a bare "timeout while waiting for state") in 8 of 400 runs.

A typed error would remove the shared variable:
```go
type fatalProbeError struct{ error }
// in the retry func: return retry.NonRetryableError(fatalProbeError{err})
var f fatalProbeError
if err != nil && !errors.As(err, &f) { /* wrap as "not confirmed deleted" */ }
```

ivanbesinovic-sysdig: Fixed in e583032.

tembleking (dismissed):

sysdig/internal/client/v2/organization.go
```diff
@@ -26,24 +34,23 @@ func (c *Client) CreateOrganizationSecure(ctx context.Context, org *Organization
...
-			err = fmt.Errorf("unable to close response body: %w", dErr)
-		}
-	}()
+	// a close failure says nothing the caller can act on, and returning it would discard a
+	// successful call: a created organization would exist server side with nothing tracking it
+	defer func() { _ = response.Body.Close() }()
 
-	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusCreated && response.StatusCode != http.StatusAccepted {
+	acknowledged := c.config.secureOrgAPIAsync && response.StatusCode == http.StatusAccepted
+	if !acknowledged && response.StatusCode != http.StatusOK && response.StatusCode != http.StatusCreated {
```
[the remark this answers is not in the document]
ivanbesinovic-sysdig: Fixed in 908eeb2 — 202 is accepted on both create and update regardless of the option, which now only decides whether an empty acknowledgement is tolerated.

tembleking (approved):

mbarbieri (approved):

